PR #99, opened 2022-06-21
[SDK-3440] Credentials Manager

frederikprijck:
### Description

Adds the Dart and Android implementation of the CredentialsManager. Also adds a dummy implementation on the Swift side to guarantee the Example application still works.

#### Expose the Android CredentialsManager using the Method Channel
Exposes the following methods from **Auth0.Android** on `credentials_manager` method channel:
  - getCredentials
  - saveCredentials
  - clearCredentials
  - hasValidCredentials

#### Expose the above MethodChannel calls using a `CredentialsManager` in Dart.
Exposes all of the above method channel calls by the means of a `DefaultCredentialsManager` in Dart, which is an our default implementation that extends `CredentialsManager`.

#### Using CredentialsManager with WebAuth 
WebAuth now uses the credentials manager by default, meaning that we store the token whenever `login` was succesful, and we remove it again once the user called `logout`.

Apart from that, WebAuth now also contains a method `credentials`, which is used to retrieve the credentials from the CredentialsManager.

#### Enable opting-out 
This introduces using the CredentialsManager by default. Users can opt out of this by calling `auth0('domain', 'clientId', useCredentialsManager: false)`.

#### Enable providing a custom credentials manager
Some users might want to use their own credentials manager.
They can do so by providing a custom credentials manager when calling `auth0('domain', 'clientId', customCredentialsManager: myCredentialsManager)`

#### Use biometrics when available
If users want to use biometrics when accessing the credentials, they can call `auth0('domain', 'clientId', useBiometrics: true)`.

- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [x] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [x] All active GitHub checks for tests, formatting, and security are passing
- [x] The correct base branch is being used, if not `master`

### DIFF
--- a/auth0_flutter/test/web_authentication_test.dart
+++ b/auth0_flutter/test/web_authentication_test.dart
@@ -125,6 +125,9 @@ void main() {
           .thenAnswer((final _) async => true);
       final mockCm = MockCredentialsManager();
 
+      when(mockCm.set(any))
+          .thenAnswer((final _) async => true);
+
       await Auth0('test-domain', 'test-clientId')
           .webAuthentication(customCredentialsManager: mockCm)
           .login(
@@ -280,6 +283,8 @@ void main() {
       when(mockedCMPlatform.clearCredentials(any))
           .thenAnswer((final _) async => true);
       final mockCm = MockCredentialsManager();
+      when(mockCm.clear())
+          .thenAnswer((final _) async => true);
 
       await Auth0('test-domain', 'test-clientId')
           .webAuthentication(customCredentialsManager: mockCm)


auth0_flutter_platform_interface/test/method_channel_credentials_manager_test.dart
```diff
@@ -0,0 +1,479 @@
...
+
+    test(
+        'throws a CredentialsManagerException when method channel throws a PlatformException',
+        () async {
+      when(mocked.methodCallHandler(any))
+          .thenThrow(PlatformException(code: '123'));
+
+      Future<bool> actual() async {
+        final result = await MethodChannelCredentialsManager().clearCredentials(
+            CredentialsManagerRequest<SaveCredentialsOptions>(
```
Widcket, line 446: Should it be using `SaveCredentialsOptions`?

auth0_flutter/lib/src/credentials_manager.dart
```diff
@@ -0,0 +1,82 @@
...
+
+/// Abstract CredentialsManager that can be used to provide a custom CredentialManager.
+abstract class CredentialsManager {
+  Future<Credentials> get({
+    final int minTtl = 0,
+    final Set<String> scopes = const {},
+    final Map<String, String> parameters = const {},
+  });
+
+  Future<void> set(final Credentials credentials);
```
Widcket, line 11: Should be `bool`.

auth0_flutter/lib/src/credentials_manager.dart
```diff
@@ -0,0 +1,82 @@
...
+
+  Future<bool> hasValidCredentials({
+    final int minTtl = 0,
+  });
+
+  Future<void> clear();
```
Widcket, line 17: Should be `bool`.

auth0_flutter/lib/auth0_flutter.dart
```diff
@@ -55,5 +47,37 @@ class Auth0 {
...
+  /// final accessToken = result.accessToken;
+  /// ```
+  ///
+  /// Uses the [DefaultCredentialsManager] by default. If you want to use your own implementation to handle credential storage, provide your own [CredentialsManager] implementation
+  /// by setting [customCredentialsManager].
+  /// In case you want to opt-out of using any [CredentialsManager] alltogether, set [useCredentialsManager] to `false`.
+  /// If you want to use biometrics or pass-phrase when using the [DefaultCredentialsManager], set [useLocalAuthentication]` to `true`.
+  /// Note however that this settings has no effect when specifying a [customCredentialsManager]
+  WebAuthentication webAuthentication({
+    final bool useCredentialsManager = true,
```
Widcket, line 69: Is still in use?

Widcket (approved):

Widcket:
Excellent work! Thanks for being patient with the reviews.

PR #108, opened 2022-07-01
Add implementation of local authentication for iOS [SDK-3460]

Widcket:
### Description

This PR build up on https://github.com/auth0/auth0-flutter/pull/104 by adding support for local authentication on iOS.

https://user-images.githubusercontent.com/5055789/177684289-c8c648d1-1d4c-4a41-a2eb-1b89e7ad6f38.MP4

### Testing

Besides adding unit tests, the changes were tested manually using an iPhone 11 Pro running iOS 15.5, with Xcode 
13.4.1 (13F100).

- [X] This change adds test coverage for new/changed/fixed functionality

### Checklist


auth0_flutter/lib/src/web_authentication.dart
```diff
@@ -86,25 +86,6 @@ class WebAuthentication {
...
-  }) async {
-    final credentials = await _credentialsManager?.get(
-      minTtl: minTtl,
-      scopes: scopes,
-      parameters: parameters,
-    );
-
-    return credentials;
-  }
-
```
frederikprijck, line 107: Is there a specific reason we are removing this method?

Widcket: Yes. Because `credentials()` will always be used in tandem with `hasValidCredentials()` (when the app starts, to check whether to show the login screen or to proceed to the main flow) and `clear()` (when logging out). Therefore it does not make sense to have a method for `credentials()` and not the others, and adding the others pretty much means wrapping the entire Credentials Manager instance. So we can simply let the developer access the Credentials Manager instance.

PR #117, opened 2022-07-12
Update `flutter_lints` to v2

Widcket:
### Description

This PR upgrades the `flutter_lints` library to v2 in all packages, and fixes the issues flagged by `flutter analyze`. Also, the rule `package_api_docs` was enabled.

### Testing

### Checklist

### DIFF
--- a/auth0_flutter/example/lib/main.dart
+++ b/auth0_flutter/example/lib/main.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:flutter_dotenv/flutter_dotenv.dart';
+// ignore: depend_on_referenced_packages
 import 'package:flutter_driver/driver_extension.dart';
 
 import 'example_app.dart';
--- a/auth0_flutter/lib/src/authentication_api.dart
+++ b/auth0_flutter/lib/src/authentication_api.dart
@@ -3,10 +3,10 @@ import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interfac
 /// An interface for calling some of the endpoints in [Auth0's Authentication API](https://auth0.com/docs/api/authentication).
 ///
 /// This class presents building blocks for doing more fine-grained authentication with Auth0 using Username and Password login. Unlike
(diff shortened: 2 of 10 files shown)


frederikprijck (approved):

auth0_flutter/example/lib/main.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:flutter_dotenv/flutter_dotenv.dart';
+// ignore: depend_on_referenced_packages
```
stevehobbsdev, line 5: Why is this exception needed?

auth0_flutter/lib/src/authentication_api.dart
```diff
@@ -3,10 +3,10 @@ import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interfac
 /// An interface for calling some of the endpoints in [Auth0's Authentication API](https://auth0.com/docs/api/authentication).
 ///
 /// This class presents building blocks for doing more fine-grained authentication with Auth0 using Username and Password login. Unlike
-/// [WebAuthentication], these do **not** use [Auth0 Universal Login](https://auth0.com/docs/authenticate/login/auth0-universal-login) (the recommended way of doing authentication),
+/// `WebAuthentication`, these do **not** use [Auth0 Universal Login](https://auth0.com/docs/authenticate/login/auth0-universal-login) (the recommended way of doing authentication),
```
stevehobbsdev, line 6: What's the reason for this change?

Widcket: The linter complains that the types referenced are not defined on this file.

auth0_flutter/example/lib/main.dart
Widcket: The linter complains that the driver package is not in the app's dependencies (it's on dev_dependencies).

frederikprijck: We could move it into the dependencies I believe, because it's actually used in the app itself so it does feel like a dependency instead of a devDependency to some extend.

PR #124, opened 2022-07-15
Move the Credentials Manager instance to `Auth0`

Widcket:
### Description

This PR moves the Credentials Manager parameter and instance from `webAuthentication()` to the `Auth0` constructor, along with the local authentication options.

#### 1. Makes it easier to use the Credentials Manager with username/password authentication (API client)

With the current implementation, if a developer wants to use username/password authentication through the API client, there is no straightforward way for them to call the built-in Credentials Manager instance to store the resulting credentials. They'd have to call `webAuthentication().credentialsManager`, even if they don't use Web Auth at all.

**Before**

```dart
final credentials = await auth0.api.login(
    usernameOrEmail: 'jane.smith@example.com',
    password: 'secret-password',
    connectionOrRealm: 'Username-Password-Authentication');

auth0.webAuthentication().credentialsManager.storeCredentials(credentials);
```

**After**

```dart
final credentials = await auth0.api.login(
    usernameOrEmail: 'jane.smith@example.com',
    password: 'secret-password',
    connectionOrRealm: 'Username-Password-Authentication');

auth0.credentialsManager.storeCredentials(credentials);
```

We'll be documenting that only Web Auth stores/renews the credentials automatically, though (in the upcoming READMEs PR).

#### 2. Makes it easier to use a single instance of the Credentials Manager

By placing the Credentials Manager instance in the top-level `Auth0` object, it becomes a matter of simply instantiating this object once and reusing it to always use the same Credentials Manager instance. Not making this the path of least resistance could lead to developers using different instances without realizing it, resulting in concurrency issues. 

**Before**

```dart
final auth0 = Auth0('domain', 'client-id');
// Every call to webAuthentication() returns a new instance
auth0.webAuthentication().credentialsManager.credentials();
```

**After**

```dart
final auth0 = Auth0('domain', 'client-id');
auth0.credentialsManager.credentials(); // Same instance
```

### Testing

### Checklist

### DIFF
--- a/auth0_flutter/lib/auth0_flutter.dart
+++ b/auth0_flutter/lib/auth0_flutter.dart
@@ -26,11 +26,29 @@ class Auth0 {
   final UserAgent _userAgent =
       UserAgent(name: 'auth0-flutter', version: version);
 
+  late final CredentialsManager _credentialsManager;
+
+  /// Secure [Credentials] store.
+  CredentialsManager get credentialsManager => _credentialsManager;
+
   /// Creates an intance of an Auth0 client with the provided [domain] and [clientId] properties.
   ///
   /// [domain] and [clientId] are both values that can be retrieved from the application in your [Auth0 Dashboard](https://manage.auth0.com).
-  Auth0(final String domain, final String clientId)
-      : _account = Account(domain, clientId);
+  /// Uses the [DefaultCredentialsManager] by default. If you want to use your own implementation to handle credential storage, provide your own [CredentialsManager] implementation
+  /// by setting [credentialsManager].
+  /// If you want to use biometrics or pass-phrase when using the [DefaultCredentialsManager], set [localAuthentication]` to an instance of [LocalAuthenticationOptions].
+  /// Note however that this setting has no effect when specifying a custom [credentialsManager].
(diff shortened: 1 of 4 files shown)


stevehobbsdev:
Random question while I'm looking over the PR description; isn't concurrency enforced at the native level, and just because we might use a different instance in the Flutter domain automatically mean that a different instance is used in the native domain? We're crossing a boundary there anyway, don't we just create the instance once on the native side regardless of what is happening in the app?

Widcket:
That is a good question, and would require hands-on testing to answer.

Widcket:
If that weren't a problem, following up with what we discussed in Slack, how could we update the current design to ensure developers are able to use the built-in instance even if they’re not using Web Auth, and do so without dealing with Web Auth APIs?

stevehobbsdev:
Ultimately I think the changes here are good, I still think the issue of where concurrency is enforced should be investigated as it **should** be enforced at the lowest level possible, not through API changes.

Happy to approve but let's give some time for @frederikprijck to have a look as well.

Widcket:
> it **should** be enforced at the lowest level possible, not through API changes.

I disagree 🙂 I think that's valid for application code, but SDKs are an entirely different matter.

PR #127, opened 2022-07-17
Add pubspec.lock to git

frederikprijck:
### Description

The flutter Orb's `install_sdk_and_pub` will try and use the lock file for the cache. Our CI is always failing on that step, even though it doesnt fail our build.

image

This PR ensures the Orb can correctly generate the cache.

However, as the [dart documentation](https://dart.dev/tools/pub/glossary#lockfile) mentions the following:

> The lockfile is generated automatically for you by pub when you run [pub get](https://dart.dev/tools/pub/cmd/pub-get), [pub upgrade](https://dart.dev/tools/pub/cmd/pub-upgrade), or [pub downgrade](https://dart.dev/tools/pub/cmd/pub-downgrade). If your package is an application package, you will typically check this into source control. For library packages, you usually won’t.

Should we acually commit this? @Widcket @stevehobbsdev ? If not, the CI step will continue to fail (but wont fail our build)

### Checklist

- [x] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [x] All active GitHub checks for tests, formatting, and security are passing
- [x] The correct base branch is being used, if not `master`


Widcket:
cc @adamjmcgrath

adamjmcgrath:
We tend to check in our lock files for our JS SDKs, even when they're not strictly needed.

Wether they're needed or not is usually quite nuanced (eg does the SDK bundle its runtime dependencies, is it transpiled) - so we just go for checking them in for everything for simplicity's sake. This is probably why the Orb works like that.

I think this PR is fine, if you really don't want to check in the lock files we could update the orb to optionally use the `package.json` (or whatever the equivalent is on flutter) or use the package.json as a fallback if the lock is not found.

adamjmcgrath (approved):

Widcket:
> we could update the orb to optionally use the package.json (or whatever the equivalent is on flutter) or use the package.json as a fallback if the lock is not found.

Note that this is a Dart library, so the orb won't have to be updated (until we modify it for Dart packages).

PR #132, opened 2022-07-20
Update SDK README and FAQ with latest API changes [SDK-3505]

Widcket:
### Description

This PR updates the SDK README and FAQ according to the current state of the API.

### Testing

### Checklist

### DIFF
--- a/auth0_flutter/README.md
+++ b/auth0_flutter/README.md
@@ -38,6 +38,7 @@ Auth0 SDK for Android / iOS Flutter apps.
 
 - Web Auth login and logout
 - Automatic storage and renewal of user's credentials
+- Support for custom Credentials Manager implementations
 - Support for the following Authentication API operations:
   + Login with username or email and password
   + Signup
@@ -232,8 +233,7 @@ final credentials = await auth0.webAuthentication().login();
 auth0_flutter will automatically store the user's credentials using the built-in [Credentials Manager](#credentials-manager) instance. You can access this instance through the `credentialsManager` property.
 
 ```dart
-final credentials =
-        await auth0.webAuthentication().credentialsManager?.credentials();
+final credentials = await auth0.credentialsManager.credentials();
 ```
 
 <details>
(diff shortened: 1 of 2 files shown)


auth0_flutter/README.md
```diff
@@ -38,6 +38,7 @@ Auth0 SDK for Android / iOS Flutter apps.
 
 - Web Auth login and logout
 - Automatic storage and renewal of user's credentials
+- Support for custom Credentials Manager implementations
```
Widcket, line 41: The abstract `CredentialsManager` was exported after all, as it is in the `auth0_flutter` package. So the respective documentation was added.

auth0_flutter/README.md
```diff
@@ -395,11 +394,7 @@ The Credentials Manager utility allows you to securely store and retrieve the us
 When the users open your app, check for valid credentials. If they exist, you can retrieve them and redirect the users to the app's main flow without any additional login steps.
 
 ```dart
-final isLoggedIn = await auth0
-        .webAuthentication()
-        .credentialsManager
-        ?.hasValidCredentials() ??
-    false;
+final isLoggedIn = await auth0.credentialsManager.hasValidCredentials();
```
frederikprijck, line 397: I do want to add that, as a developer, seeing code like this I would expect `auth0.api.login` to result in `hasValidCredentials` to be `true`, which is not true.

Widcket: Thanks for the observations. We currently are referencing that in this README, in the following instances:
- Documentation of `auth0.api.login()`: snippet includes call to `auth0.credentialsManager.storeCredentials()` afterward. https://github.com/auth0/auth0-flutter/pull/132/files#diff-300bb421482a8868624c82300a9e4332736d0cd6684c32fccf6ccd2e60bafa3eR483
- Note in the **Credentials** section: https://github.com/auth0/auth0-flutter/blob/867948bea55804940704656e29794c729eeec9ab/auth0_flutter/README.md#credentials-manager
- Documentation on custom implementations: https://github.com/auth0/auth0-flutter/pull/132/files#diff-300bb421482a8868624c82300a9e4332736d0cd6684c32fccf6ccd2e60bafa3eR418

What else do you suggest we can do to better convey this?

Issue #168, opened 2022-09-23
Support enrollMfa and requestMfa using API


g0josh:
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

Please add support for enrolling and requesting MFA challenges using API. 

### Describe the ideal solution

- enrollMfa for doing what `https://YOUR_DOMAIN/mfa/associate` does, as of now I'd have to use the REST endpoint
- requestMfa for issuing a challenge if I have my authenticator enrolled. Something like what [Auth0.swift does](https://auth0.github.io/Auth0.swift/Protocols/Authentication.html#/s:5Auth014AuthenticationP20multifactorChallenge8mfaToken5types15authenticatorIdAA7RequestVyAA0D0VAA0B5ErrorVGSS_SaySSGSgSSSgtF:~:text=multifactorChallenge(mfaToken%3Atypes%3AauthenticatorId%3A))
- loginMfa for logging in with the authenticator code received, [Auth0.swift example](https://auth0.github.io/Auth0.swift/Protocols/Authentication.html#/s:5Auth014AuthenticationP20multifactorChallenge8mfaToken5types15authenticatorIdAA7RequestVyAA0D0VAA0B5ErrorVGSS_SaySSGSgSSSgtF:~:text=login(withOOBCode%3AmfaToken%3AbindingCode%3A))

### Alternatives and current workarounds

[...]

PR #172, opened 2022-10-14
afpi: Added support for `loginWithOtp` 

wferem1:

### 📋 Changes

* Expanded the plugin with the `loginWithOtp` method, both for iOS and Android.
* Added a convenience method in `ApiException` to get the `mfaToken` from the `ApiException` when `mfaIsRequired` is true

If you receive an `ApiException` with `mfaIsRequired` set to true, you can now prompt the user for a OTP code and call `Auth0.api.loginWithOtp` with the `mfaToken` present in the `ApiException` and the user's otp.

### 📎 References
This PR solves part of the features requested in #168


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
```diff
@@ -3,32 +3,33 @@ package com.auth0.auth0_flutter
...
+            "isNetworkError" to ex.isNetworkError,
+            "isBrowserAppNotAvailable" to ex.isBrowserAppNotAvailable,
+            "isPKCENotAvailable" to ex.isPKCENotAvailable,
+            "isInvalidAuthorizeURL" to ex.isInvalidAuthorizeURL,
+            "isInvalidConfiguration" to ex.isInvalidConfiguration,
+            "isCanceled" to ex.isCanceled,
+            "isPasswordLeaked" to ex.isPasswordLeaked,
+            "isLoginRequired" to ex.isLoginRequired,
+        ))
+        if (ex.getValue("mfa_token") != null) { put("mfa_token", ex.getValue("mfa_token")!!) }
```
Widcket, line 31: Let's use camelCase for the key.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
```diff
@@ -3,31 +3,33 @@ package com.auth0.auth0_flutter
...
-            "isBrowserAppNotAvailable" to this.isBrowserAppNotAvailable,
-            "isPKCENotAvailable" to this.isPKCENotAvailable,
-            "isInvalidAuthorizeURL" to this.isInvalidAuthorizeURL,
-            "isInvalidConfiguration" to this.isInvalidConfiguration,
-            "isCanceled" to this.isCanceled,
-            "isPasswordLeaked" to this.isPasswordLeaked,
-            "isLoginRequired" to this.isLoginRequired,
-        )
-    )
+    var ex = this;
```
Widcket, line 6: Let's use a non-abbreviated word for better readability, e.g. `exception`.

auth0_flutter_platform_interface/lib/src/method_channel_auth0_flutter_auth.dart
```diff
@@ -15,6 +16,7 @@ import 'user_profile.dart';
 
 const MethodChannel _channel = MethodChannel('auth0.com/auth0_flutter/auth');
 const String authLoginMethod = 'auth#login';
+const String authLoginWithOtpMethod = 'auth#login_otp';
```
Widcket, line 19: All the other method keys use camelCase, please use it for this new key as well.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
```diff
@@ -44,6 +44,7 @@ class LoginApiRequestHandler : ApiRequestHandler {
         if (args.getOrDefault("parameters", null) is HashMap<*, *>) {
             loginBuilder.addParameters(args["parameters"] as Map<String, String>)
         }
+        loginBuilder.validateClaims();
```
Widcket, line 47: I don't think we should enable this without also providing a way to customize the ID Token validation (e.g. the issuer -> `withIdTokenVerificationIssuer ()` in Auth0.Android). Since this would be outside the scope of this PR, better leave this for another PR.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginOtpApiRequestHandler.kt
```diff
@@ -0,0 +1,69 @@
...
+        val args = request.data
+
+        assertHasProperties(listOf("mfaToken", "otp"), args)
+
+        val loginBuilder = api
+            .loginWithOTP(
+                args["mfaToken"] as String,
+                args["otp"] as String,
+            )
+        loginBuilder.validateClaims();
```
Widcket, line 35: Same as above.

auth0_flutter/ios/Classes/AuthAPI/AuthAPILoginWithOtpMethodHandler.swift
```diff
@@ -0,0 +1,30 @@
...
+        guard let otp = arguments[Argument.otp] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing(Argument.otp.rawValue)))
+        }
+        guard let mfaToken = arguments[Argument.mfaToken] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing(Argument.mfaToken.rawValue)))
+        }
+
+        client
+            .login(withOTP: otp,
+                  mfaToken: mfaToken)
```
Widcket, line 22:
```suggestion
            .login(withOTP: otp, mfaToken: mfaToken)
```

auth0_flutter/lib/src/authentication_api.dart
```diff
@@ -73,6 +73,31 @@ class AuthenticationApi {
         parameters: parameters,
       )));
 
+  /// Authenticates the user using a [mfaToken] and a [otp], after [login] returned with an [ApiException] with [ApiException.isMultifactorRequired] set to true.
```
Widcket, line 76:
```suggestion
  /// Authenticates the user using a [mfaToken] and an [otp], after [login] returned with an [ApiException] with [ApiException.isMultifactorRequired] set to true.
```

auth0_flutter/lib/src/authentication_api.dart
```diff
@@ -73,6 +73,31 @@ class AuthenticationApi {
...
+  ///
+  /// ## Endpoint docs
+  /// https://auth0.com/docs/api/authentication#verify-with-one-time-password-otp-
+  ///
+  /// ## Usage example
+  ///
+  /// ```dart
+  /// final result = await auth0.api.loginWithOtp({
+  ///   otp: '123456',
+  ///   mfaToken: 'receivedMfaToken'
```
Widcket, line 88:
```suggestion
  ///   mfaToken: 'received_mfa_token'
```

Widcket: Let's continue the convention followed by other API docs from the same package.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
```diff
@@ -3,31 +3,33 @@ package com.auth0.auth0_flutter
...
+            "isNetworkError" to ex.isNetworkError,
+            "isBrowserAppNotAvailable" to ex.isBrowserAppNotAvailable,
+            "isPKCENotAvailable" to ex.isPKCENotAvailable,
+            "isInvalidAuthorizeURL" to ex.isInvalidAuthorizeURL,
+            "isInvalidConfiguration" to ex.isInvalidConfiguration,
+            "isCanceled" to ex.isCanceled,
+            "isPasswordLeaked" to ex.isPasswordLeaked,
+            "isLoginRequired" to ex.isLoginRequired,
+        ))
+        if (ex.getValue("mfa_token") != null) { put("mfa_token", ex.getValue("mfa_token")!!) }
```
Widcket, line 31: Let's use camelCase for the key.

auth0_flutter_platform_interface/lib/src/auth/api_exception.dart
```diff
@@ -35,6 +35,10 @@ class ApiException extends Auth0Exception {
 
   bool get isMultifactorRequired =>
       _errorFlags.getBooleanOrFalse('isMultifactorRequired');
+  String? get mfaToken =>
+      details.containsKey('mfa_token') && details['mfa_token'] is String
```
Widcket, line 39: Same: let's use camelCase for the key.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
wferem1: I'm not the most experienced with kotlin, but the `buildMap` method was my solution to have an optional key-value pair in the map, instead of changing the signature of the map from <String, Any> to <String, Any?>. In this method referencing `this` would point to the map instead of the exception, which is why I had to introduce this variable.
If you have a better method, feel free to let me know

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
wferem1: on iOS, this key is already present in `Auth0APIError.details` with key `mfa_token`, which is why i chose to do have the same key on the android side. I could replace this key in `auth0_flutter/ios/Classes/Extensions.swift:47`,

Widcket:
> Since I made changes in the platform_interface, the plugin would need these changes, I could add a dependency override,

We typically make the dependency override when developing, but change it back before releasing. Please change it so we can see if the CI passes.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
Widcket: No worries, I'm no Kotlin expert myself either. I'll ask @poovamraj to review the Android bits once we're close to merging.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
Widcket: Do you mean here https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/ios/Classes/Extensions.swift#L47?

wferem1: Yes, self.info` already contains the mfa token, but with `mfa_token` as key, I started with `mfaToken` until I tried testing the plugin on an iOS device.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
wferem1: I removed these

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/AuthenticationExceptionExtensions.kt
Widcket: Yes, that's right.

Widcket:
@poovamraj could you please review the Kotlin/Android bits?

Widcket:
@wferem1 could you please correct the title? Once you edit it, the corresponding CI check will run again.

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginWithOtpApiRequestHandlerTest.kt
```diff
@@ -0,0 +1,170 @@
...
+        assertThat(
+            exception.message,
+            equalTo("Required property 'otp' is not provided.")
+        )
+    }
+
+    @Test
+    fun `should throw when missing mfaToken`() {
+        val options =
+            hashMapOf("otp" to "test-otp")
```
poovamraj, line 52:
```suggestion
        val options = hashMapOf("otp" to "test-otp")
```

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginWithOtpApiRequestHandlerTest.kt
```diff
@@ -0,0 +1,170 @@
...
+import org.robolectric.RobolectricTestRunner
+import java.text.SimpleDateFormat
+import java.util.*
+
+@RunWith(RobolectricTestRunner::class)
+class LoginWithOtpApiRequestHandlerTest {
+    @Test
+    fun `should throw when missing otp`() {
+        val options =
+            hashMapOf("mfaToken" to "test-mfa-token")
```
poovamraj, line 28:
```suggestion
        val options = hashMapOf("mfaToken" to "test-mfa-token")
```

poovamraj (commented):
Made a few small suggestions. Should be easy.

Widcket:
@wferem1 are you still interested in this?

wferem1:
@Widcket 
Sorry, I was a bit busy, and because of security reasons, I decided to go back to the web auth login flow. 
The reason I wanted to support this, was that the browser on the device stores cookies, so that the OTP in the web auth flow is only asked once every 30 days.

However, allowing these login methods, seems less secure than the PKCE flow. 

I can try and take a look at the remarks in my free time, as it seems like a good addition for those who want to use the OTP method. But it's not an urgent request for me anymore.

auth0_flutter/lib/src/authentication_api.dart
```diff
@@ -73,6 +73,31 @@ class AuthenticationApi {
         parameters: parameters,
       )));
 
+  /// Authenticates the user using a [mfaToken] and an [otp], after [login] returned with an [ApiException] with [ApiException.isMultifactorRequired] set to true.
```
Widcket, line 76:
```suggestion
  /// Authenticates the user using a [mfaToken] and an [otp], after [login] returned with an [ApiException] with [ApiException.isMultifactorRequired] set to `true`.
```

auth0_flutter_platform_interface/lib/src/auth0_flutter_auth_platform.dart
```diff
@@ -30,6 +31,11 @@ abstract class Auth0FlutterAuthPlatform extends PlatformInterface {
     throw UnimplementedError('authLogin() has not been implemented');
   }
 
+  Future<Credentials> loginWithOtp(
+      final ApiRequest<AuthLoginWithOtpOptions> request) {
+    throw UnimplementedError('authLogin() has not been implemented');
```
Widcket, line 36:
```suggestion
    throw UnimplementedError('authLoginWithOtp() has not been implemented');
```

Widcket (approved):

Issue #178, opened 2022-10-28
NoSuchMethodError: No interface method getOrDefault


tijee:
### Checklist

- [X] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

From an error report in Sentry:

```
java.lang.NoSuchMethodError: No interface method getOrDefault(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object; in class Ljava/util/Map; or its super classes (declaration of 'java.util.Map' appears in /system/framework/core-libart.jar)
    at com.auth0.auth0_flutter.g.d.a.b(LoginWebAuthRequestHandler.kt:3)
    at com.auth0.auth0_flutter.c.onMethodCall(Auth0FlutterWebAuthMethodCallHandler.kt:3)
    at io.flutter.c.a.j$a.a(MethodChannel.java:2)
    at io.flutter.embedding.engine.f.c.i(DartMessenger.java:2)
    at io.flutter.embedding.engine.f.c.j(DartMessenger.java:2)
    at io.flutter.embedding.engine.f.c.k(DartMessenger.java)
    at io.flutter.embedding.engine.f.a.run(lambda)
    at android.os.Handler.handleCallback(Handler.java:739)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at android.os.Looper.loop(Looper.java:148)
    at android.app.ActivityThread.main(ActivityThread.java:5421)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
[...]

PR #179, opened 2022-10-29
af: Replace usages of `getOrDefault()`

Widcket:

### 📋 Changes

This PR removes or replaces all usages of the `getOrDefault()` on the Kotlin code, as this method is only available on Android API 24+: https://developer.android.com/reference/java/util/Map#getOrDefault(java.lang.Object,%20V)

For the cases where there was a non-null default value, the Elvis operator was used to provide the default value.

For the cases where the default value was simply `null`, there was no need to do anything else as the map already returns `null` if the value is not present:

Screen Shot 2022-10-28 at 20 38 33

Also, an unused import was removed.

### 📎 References

Fixes https://github.com/auth0/auth0-flutter/issues/178


tijee:
Hi @Widcket, any idea when it will be released?

Widcket:
cc @stevehobbsdev

stevehobbsdev:
[1.0.2](https://pub.dev/packages/auth0_flutter) has now been published 🎉

PR #199, opened 2023-02-23
Upgrade Flutter used for build, and fix linter warnings

stevehobbsdev:
- N/A All new/changed/fixed functionality is covered by tests (or N/A)
- N/A I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

- Upgrade version of Flutter used in build to 3.7.4
- Fix linter warnings, mostly around lines > 80 chars in comments. A couple of tests violated this, thought it better to ignore the file rather than obfusticate the test name for this

Primary motivation was that Ship CLI fails when running `flutter analyze` (now returns a non-zero exit code even for warnings). Plus, good to keep up to date with latest Flutter anyway, and linter warnings should be fixed before they get out of hand.

### DIFF
--- a/auth0_flutter/analysis_options.yaml
+++ b/auth0_flutter/analysis_options.yaml
@@ -17,7 +17,7 @@ analyzer:
     unawaited_futures: error
     parameter_assignments: warning
     included_file_warning: warning
-    todo: ignore # 'warning' after EA
+    todo: warning
 
 linter:
   rules:
--- a/auth0_flutter/example/analysis_options.yaml
+++ b/auth0_flutter/example/analysis_options.yaml
@@ -26,7 +26,7 @@ analyzer:
     unawaited_futures: error
     parameter_assignments: warning
     included_file_warning: warning
-    todo: ignore # 'warning' after EA
+    todo: warning
 
 linter:
   # The lint rules applied to this project can be customized in the
--- a/auth0_flutter_platform_interface/analysis_options.yaml
+++ b/auth0_flutter_platform_interface/analysis_options.yaml
@@ -17,7 +17,7 @@ analyzer:
     unawaited_futures: error
     parameter_assignments: warning
     included_file_warning: warning
-    todo: ignore # 'warning' after EA
+    todo: warning
 
 linter:
   rules:


auth0_flutter/analysis_options.yaml
```diff
@@ -52,7 +52,7 @@ linter:
     join_return_with_assignment: true
     leading_newlines_in_multiline_strings: true
     library_private_types_in_public_api: true
-    lines_longer_than_80_chars -> packages only: true
```
stevehobbsdev, line 55: Flutter CLI doesn't know what `packages only` does any more, and I can't find much on the web about it.

auth0_flutter/example/lib/api_card.dart
```diff
@@ -6,7 +6,7 @@ import 'constants.dart';
 
 class ApiCard extends StatefulWidget {
   final Future<void> Function(String usernameOrEmail, String password) action;
-  const ApiCard({required final this.action, final Key? key}) : super(key: key);
```
stevehobbsdev, line 9: Fixes `unnecessary final` warnings, as these are implicitly `final`.

auth0_flutter/analysis_options.yaml
Widcket: That's a comment by me, which I forgot to remove when I uncommented the line. My bad 🤦🏼‍♀️

auth0_flutter/example/analysis_options.yaml
```diff
@@ -73,7 +73,7 @@ linter:
     join_return_with_assignment: true
     leading_newlines_in_multiline_strings: true
     library_private_types_in_public_api: true
-    lines_longer_than_80_chars -> packages only: true
+    lines_longer_than_80_chars: true
```
Widcket, line 76: Could you please change `todo` to `warning` on L29? Also for the  other YAML files.

auth0_flutter/analysis_options.yaml
stevehobbsdev: All good, I had thought it was some kind of special hidden syntax we were using 😄

auth0_flutter/example/analysis_options.yaml
stevehobbsdev: Done in 110bd0c

Widcket (approved):
Thank you!

PR #203, opened 2023-02-23
[SDK-3984] Support login on web platform

stevehobbsdev:

### 📋 Changes

This PR presents an initial implementation of the Flutter Web with JS interop and Auth0 SPA SDK. It supports:

* login using `loginWithRedirect`
* `checkSession`
* Automatic use of `handleRedirectCallback` on load

Things that are pending in forthcoming PRs:

* all options for client set up and login (currently only a subset of things like [`AuthorizeParams`](https://auth0.github.io/auth0-spa-js/interfaces/RedirectLoginOptions.html#authorizationParams) are supported until the API matures)
* test suite (do once the API is stable)
* support for getting tokens
* logout

**Usage** 

```dart
// inside some initializer func like 'initState':
@override
void initState() {
  super.initState();

  auth0Web =
      Auth0Web(dotenv.env['AUTH0_DOMAIN']!, dotenv.env['AUTH0_CLIENT_ID']!);

  // call to initialize the client and auto-handle callback or check session
  auth0Web.onLoad();
}

// In a login button click handler..
Future<void> webAuthLogin() async {
  String output = '';

  // Redirect the user agent to the login page
  auth0Web.loginWithRedirect(redirectUri: 'http://localhost:3000');
}
```

This PR also includes a refactoring of `WebAuthLoginOptions` to create a base type `LoginOptions` that can also be used by the web platform. The former extends the latter and adds the two mobile platform properties `useEphemeralSession` and `scheme` that are not applicable for the web.

### 🎯 Testing

The example app has been updated to support login when using the Web platform. To run:

```cli
flutter run -d web-server --web-port 3000
```

The example app will use the same domain & client ID as defined in `.env`.


auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -1,2 +1,86 @@
...
+  external String? redirect_uri;
+
+  // https://github.com/dart-lang/sdk/issues/38445
+  // Workaround of setting property names doesn't seem to work. Tried:
+  // @JS('redirect_uri')
+  // external String? get redirectUri;
+
+  // @JS('redirect_uri')
+  // external set redirectUri(final String? redirectUri);
+}
```
stevehobbsdev, line 32: Most if not all the options will be supported here eventually, but having a couple to start off with helps to see the creation patterns. e.g. it took a while to see if we could support `redirectUri` in the Dart interop API but `redirect_uri` in JS-land. tl;dr is, we can't (yet).

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -1,2 +1,86 @@
...
+    final key = keys[i] as String;
+    final value = getProperty(obj, key) as dynamic;
+
+    if (value != null) {
+      setProperty(output, key, value);
+    }
+  }
+
+  return output as T;
+}
```
stevehobbsdev, line 50: Some mapping between the keys to strip out values that were `null` is necessary, otherwise the nulls get sent through the URL to the IdP, which is undesirable. The SPA SDK doesn't strip these out.

This is needed as Dart JS interop treats `null` and `undefined` as completely identical. Consider this example:

```dart
final String? redirectUri = null;

client.loginWithRedirect(RedirectLoginOptions(
  authorizationParams: AuthorizationParams(
    redirect_uri: redirectUri)));

// redirect_uri is `null` in the URL
```

Without some mapping function that executes in JS land, an alternative would be to do two different calls depending on whether `redirectUri` is null or not, which doesn't scale as this applies to all optional parameters.

Or, use `Map<string, string>` in the API, in which case we lose strong typing but also in JS land this equates to using a Dart Map type, which is different from a plain JS object, so the JS interop API doesn't line up with reality (**must** use classes annotated with `@anonymous`).

stevehobbsdev: Again, not precious on the name of the function, happy to hear suggestions.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -1,13 +1,41 @@
...
+        Auth0ClientOptions(domain: account.domain, clientId: account.clientId));
+
+    final search = window.location.search;
+
+    if (search?.contains('state=') == true &&
+        (search?.contains('code=') == true ||
+            search?.contains('error=') == true)) {
+      await promiseToFuture<void>(client.handleRedirectCallback());
+    } else {
+      await promiseToFuture<void>(client.checkSession());
```
stevehobbsdev, line 31: Just a note to say that if you want to read the values from resolved promises, you **must** wrap it in `promiseToFuture` to convert it to a Dart Future. Otherwise you just get a pending Promise object back.

auth0_flutter/example/web/index.html
```diff
@@ -38,6 +38,7 @@
   </script>
   <!-- This script adds the flutter initialization JS code -->
   <script src="flutter.js" defer></script>
+  <script src="sdk-2.0.js" defer></script>
```
Widcket, line 41: Can we use the CDN here?

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -1,2 +1,79 @@
...
+
+@JS('auth0')
+library auth0;
+
+import 'dart:js_util';
+
+import 'package:js/js.dart';
+
+@JS()
+@staticInterop
```
Widcket, line 14: Note that this is experimental, and we should use the current form of interop: `For now, static interop remains experimental and in development. We may make breaking changes in the future. We'll update this text when the new interop model matures and is considered stable.`. See https://pub.dev/packages/js#static-interop

auth0_flutter/lib/src/web/js_interop.dart
Widcket: Does the workaround work when not using static interop?

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -0,0 +1,25 @@
...
+  final UserAgent _userAgent =
+      UserAgent(name: 'auth0-flutter', version: version);
+
+  Auth0Web(final String domain, final String clientId)
+      : _account = Account(domain, clientId);
+
+  Future<Credentials?> onLoad() async {
+    await Auth0FlutterWebPlatform.instance.initialize(_account);
+
+    // Get credentials here and return them on load?
```
Widcket, line 16: No, I don't think that'd be a good idea.

auth0_flutter_platform_interface/lib/src/web-auth/web_auth_login_options.dart
```diff
@@ -1,40 +1,33 @@
...
   final String? scheme;
 
   WebAuthLoginOptions(
-      {this.idTokenValidationConfig,
-      this.audience,
-      this.scopes = const {},
-      this.redirectUrl,
-      this.organizationId,
-      this.invitationUrl,
+      {final IdTokenValidationConfig? idTokenValidationConfig,
```
Widcket, line 9: I think these can be defined using `super`:

Screenshot 2023-02-23 at 16 16 02

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -0,0 +1,25 @@
...
+
+    return null;
+  }
+
+  Future<void> loginWithRedirect(
+          {final String? audience, final String? redirectUri}) =>
```
Widcket, line 22: Should we use `redirectUrl` instead?

auth0_flutter/lib/auth0_flutter_web.dart
Widcket, line 16: Could the consumer use the result to check if the user is logged in?

Widcket: If so, I'd say go for it.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -1,13 +1,41 @@
...
   }
 
-  // methods, e.g. loginWithRedirect()
-  // here they should call the interop methods, that in turn call the JS methods
+  late Auth0Client client;
+
+  @override
+  Future<void> initialize(final Account account,
+      {final AuthorizationParams? authorizationParams}) async {
+    client = Auth0Client(
```
Widcket, line 21: Should we create a new instance only when there isn' one already?

auth0_flutter/example/lib/example_app.dart
```diff
@@ -21,29 +23,40 @@ class _ExampleAppState extends State<ExampleApp> {
...
-
-      setState(() {
-        _isLoggedIn = true;
-      });
-    } on WebAuthenticationException catch (e) {
-      output = e.toString();
+    String output = '';
+
+    if (kIsWeb) {
+      auth0Web.loginWithRedirect(redirectUri: 'http://localhost:3000');
```
Widcket, line 46: Should this have an `await`, and be all wrapped in the same try/catch?

Widcket: Also, what happens with the errors returned from JS land?

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -1,2 +1,79 @@
...
+  // @JS('redirect_uri')
+  // external String? get redirectUri;
+
+  // @JS('redirect_uri')
+  // external set redirectUri(final String? redirectUri);
+}
+
+/// Rebuilds the input object, omitting values that are null
+@JS()
+T stripNulls<T extends Object>(final T obj) {
```
Widcket, line 36: Is the `@JS()` annotation necessary here?

auth0_flutter/lib/src/web/js_interop.dart
stevehobbsdev: Unfortunately not. In fact it was this problem that prompted me to try static interop to see if it enabled it.

auth0_flutter/example/web/index.html
stevehobbsdev: We can indeed. My thinking here was just to include it in the package to make it more robust, but considering CDN availability, caching, etc, we're probably better off including it from the CDN 👍🏻

auth0_flutter/lib/src/web/js_interop.dart
stevehobbsdev: Thanks, I missed that. I'll swap it back

stevehobbsdev: Although we'll want to move back to it at some stage once it's stable.

auth0_flutter_platform_interface/lib/src/web-auth/web_auth_login_options.dart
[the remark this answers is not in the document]
stevehobbsdev: Great!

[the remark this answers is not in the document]
stevehobbsdev: I actually fixed this in 68a105f when I merged `beta` and fixed up some conflicts. Much nicer

auth0_flutter/example/web/index.html
```diff
@@ -52,7 +53,7 @@
             appRunner.runApp();
           });
         }
-      });
+      });
```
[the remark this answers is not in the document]
stevehobbsdev: Fixed it locally 👍🏻

auth0_flutter/lib/auth0_flutter_web.dart
stevehobbsdev: Yes, I suppose the idea would be if the returned `Credentials` is not null, the user could be considered logged in. The use case I imagined would be if you were to initialize it in `initState`, the developer could then set up their initial app state accordingly:

```dart
@override
void initState() {
  super.initState();

  auth0Web =
      Auth0Web(dotenv.env['AUTH0_DOMAIN']!, dotenv.env['AUTH0_CLIENT_ID']!);

  // call to initialize the client and auto-handle callback or check session
  auth0Web.onLoad().then((creds) {
    if (creds != null) {
      // set "logged in" state here
    }
  });
}
```

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
stevehobbsdev: The more I play with this, the more I don't think it's possible or at least intuitive.

I can do:

```dart
if (client == null) {
  client = Auth0Client(Auth0ClientOptions(
      domain: account.domain, clientId: account.clientId));
}
```

But because `client` is non-optional, the block never runs - I guess the compiler is optimizing out the statement as it knows it'll never be executed. Same if I try:

```dart
client ??= Auth0Client(Auth0ClientOptions(
      domain: account.domain, clientId: account.clientId));
```

And actually the linter warns me about this one.

We could track some internal state though to tell whether the client has been initialized. We could also type it as `Auth0Client?` and manage possible null references everywhere, but that's a bit more of a pain.

auth0_flutter/example/lib/example_app.dart
stevehobbsdev: > Should this have an await, and be all wrapped in the same try/catch?

Yes, probably should.

> Also, what happens with the errors returned from JS land?

They _should_ get surfaced in the same way. Something to test, though.

auth0_flutter/lib/src/web/js_interop.dart
stevehobbsdev: Doesn't look like it! Will remove 👍🏻

stevehobbsdev: Removed it. I also moved it out into the plugin file as a private method, keeping the interop file just for interop stuff.

auth0_flutter/lib/src/web/js_interop.dart
stevehobbsdev: Fixed in https://github.com/auth0/auth0-flutter/pull/203/commits/1355315f9d09a217b5a6c4b93fa4a23e19ab8b08

auth0_flutter/example/web/index.html
stevehobbsdev: Fixed in https://github.com/auth0/auth0-flutter/pull/203/commits/573e301b4ccacf1cc4d078a19889c51ff1cd1c3a

auth0_flutter_platform_interface/lib/src/web-auth/web_auth_login_options.dart
Widcket: Yes, my bad 🤦🏼

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
Widcket: Let's leave it this way then.

Widcket (approved):

PR #207, opened 2023-02-27
Add `credentials()` and `hasValidCredentials()` [SDK-3997]

Widcket:

### 📋 Changes

This PR adds support for fetching the credentials, and for checking if there are any valid credentials stored. The following public methods were added:

- `credentials()` -> calls `getTokenSilently()` from the SPA SDK.
- `hasValidCredentials()` -> calls `isAuthenticated()` from the SPA SDK.

### 🎯 Testing

The changes were tested manually. Unit tests will be added in a future PR.

https://user-images.githubusercontent.com/5055789/221452162-4a7a8972-7c2e-4d9f-a109-776c204b4f9d.mov

### DIFF
--- a/auth0_flutter/lib/src/web/credentials_extension.dart
+++ b/auth0_flutter/lib/src/web/credentials_extension.dart
@@ -1,0 +1,21 @@
+import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interface.dart';
+
+import 'js_interop.dart';
+import 'jwt_decode.dart';
+import 'user_profile_extension.dart';
+
+extension CredentialsExtension on Credentials {
+  static Credentials fromWeb(final WebCredentials webCredentials) {
+    final expiresIn = webCredentials.expires_in as int;
+    final expiresAt = DateTime.now().add(Duration(seconds: expiresIn));
+    final claims = JWT.decode(webCredentials.id_token);
+    final user = UserProfileExtension.fromClaims(claims);
+
+    return Credentials(
+        idToken: webCredentials.id_token,
+        accessToken: webCredentials.access_token,
+        expiresAt: expiresAt,
+        user: user,
+        tokenType: 'Bearer');
+  }
+}
--- a/auth0_flutter/lib/src/web/jwt_decode.dart
+++ b/auth0_flutter/lib/src/web/jwt_decode.dart
@@ -1,0 +1,28 @@
+import 'dart:convert';
+
+import 'jwt_decode_exception.dart';
+
+class JWT {
+  static Map<String, dynamic> decode(final String jwt) {
+    final parts = jwt.split('.');
+
+    if (parts.length != 3) {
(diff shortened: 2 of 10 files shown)


auth0_flutter/lib/src/web/jwt_decode_exception.dart
```diff
@@ -0,0 +1,26 @@
...
+  const JWTDecodeException.invalidJSON(final String value)
+      : this(JWTDecodeException._invalidJSON,
+            'Failed to parse JSON from Base64URL value $value.');
+
+  // When the JWT doesn't have the required amount of parts (header, body, and
+  // signature).
+  const JWTDecodeException.invalidPartCount(final String jwt, final int parts)
+      : this(JWTDecodeException._invalidPartCount,
+            'The JWT $jwt has $parts parts when it should have 3 parts.');
+}
```
Widcket, line 26: These are the same messages and error cases as the ones in JWTDecode.swift: https://github.com/auth0/JWTDecode.swift/blob/master/JWTDecode/JWTDecodeError.swift

auth0_flutter/lib/src/web/credentials_extension.dart
```diff
@@ -0,0 +1,21 @@
...
+    final expiresAt = DateTime.now().add(Duration(seconds: expiresIn));
+    final claims = JWT.decode(webCredentials.id_token);
+    final user = UserProfileExtension.fromClaims(claims);
+
+    return Credentials(
+        idToken: webCredentials.id_token,
+        accessToken: webCredentials.access_token,
+        expiresAt: expiresAt,
+        user: user,
+        tokenType: 'Bearer');
```
stevehobbsdev, line 19: I believe we get `tokenType` from the `getTokenSilently` detailed response - should we use that, or hard code it?

auth0_flutter/lib/src/web/jwt_decode.dart
```diff
@@ -0,0 +1,28 @@
...
+    } catch (error) {
+      throw JWTDecodeException.invalidBase64URL(parts[1]);
+    }
+
+    try {
+      return jsonDecode(stringPayload) as Map<String, dynamic>;
+    } catch (error) {
+      throw JWTDecodeException.invalidJSON(parts[1]);
+    }
+  }
```
stevehobbsdev, line 27: We can already get the decoded token from the SDK using `getIdTokenClaims` - do we need to decode it manually here? 

https://auth0.github.io/auth0-spa-js/classes/Auth0Client.html#getIdTokenClaims

Widcket: Not sure we can represent custom claims (https://github.com/auth0/auth0-spa-js/blob/master/src/global.ts#L578) without using a `Map<String, dynamic>`, hence the custom decoding.

stevehobbsdev (approved):

PR #216, opened 2023-03-08
[SDK-4000] JS interop tests

stevehobbsdev:

### 📋 Changes

This PR adds tests for `Auth0Web` with the following scenarios:

* Initalization
* Login using redirect
* Handling the redirect callback
* Checking session

Also:

* Implements the wrapper code around `credentials` that first checks `hasValidCredentials` (was needed for tests to be working)
* Adds a proxy between the platform instance and the JS interop `Auth0Client` object, so that it can be mocked (we had great difficulty mocking `Auth0Client` with the interop between `Promise` and `Future`)
* Injects a function that returns `window.location.search` - setting this in a test interfered with the test runner, as the tests run in a browser instance


Widcket (approved):
Brilliant! You extracted out the things that were problematic to test, so only these could be mocked, and nothing else. This allows us to use the built-in `promiseToFuture()`, without having to declare a `Promise` type, keeping the solution simple and clean. The kind of thing that makes you go "Of course!" when you see it.
Well done!

PR #223, opened 2023-03-10
sdk 4001/logout

stevehobbsdev:

### 📋 Changes

This PR adds logout support for the web platform. Additional comments made below on specific code points.

### 📎 References

`SDK-4001` in Jira.

### 🎯 Testing

The example app was modified to support using web logout as well as on mobile.


auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -1,12 +1,15 @@
...
+
 import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interface.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+
 import 'auth0_flutter_web_platform_proxy.dart';
 import 'extensions/client_options_extensions.dart';
 import 'extensions/credentials_extension.dart';
-import 'js_interop.dart';
+import 'extensions/logout_options.extension.dart';
+import 'js_interop.dart' as interop;
```
stevehobbsdev, line 11: As we now have `LogoutOptions` in both the platform interface and JS interop, I prefixed the interop types. Kind of useful to distinguish them anyway.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -1,12 +1,15 @@
...
+import 'js_interop_utils.dart';
```
stevehobbsdev, line 12: Moved `stripNulls` out to a utility class so that they could also be used in extensions.

auth0_flutter/lib/src/web/auth0_flutter_plugin_stub.dart
```diff
@@ -30,4 +30,9 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
     throw UnsupportedError(
         'hasValidCredentials is only supported on the web platform');
   }
+
+  @override
+  Future<void> logout(final LogoutOptions? options) {
+    throw UnsupportedError('logout is only supported on the web platform');
+  }
```
stevehobbsdev, line 37: This is a little weird and one case where sharing the existing platform implementation might have been useful.

auth0_flutter/test/web/auth0_fluter_web_test.dart
```diff
@@ -153,4 +153,15 @@ void main() {
...
+  test('logout is called and succeeds', () async {
+    when(mockClientProxy.logout(any)).thenAnswer((final _) => Future.value());
+    await auth0.logout(federated: true, returnToUrl: 'http://returnto.url');
+
+    final params =
+        verify(mockClientProxy.logout(captureAny)).captured.first.logoutParams;
+
+    expect(params.federated, true);
+    expect(params.returnTo, 'http://returnto.url');
+  });
```
Widcket, line 166: Should we add a failing test? It'll have to be changed when we add our new exception, but at least it will be in place.

Widcket: Also, should we test that the nulls are stripped, like in "loginWithRedirect strips options that are null"?

stevehobbsdev: > Should we add a failing test?

I've added one now in [ab3524b](https://github.com/auth0/auth0-flutter/pull/223/commits/ab3524b9edb147e1daae8939a360b9ee285edd54) that can be rewritten later 👍🏻 

> Also, should we test that the nulls are being stripped, like in "loginWithRedirect strips options that are null"?

Yes I agree that's a good test, and in fact started down this road as part of the login work but decided to revisit it. There's no easy way to distinguish between something that's `null` and something that's not been set. If you take this example:

```dart
final params = verify(mockClientProxy.loginWithRedirect(captureAny))
    .captured
    .first
    .authorizationParams;

expect(params.audience, null);
```

`params.audience` will be `null` if it hasn't been set or if it's actually been set to `null`. To test `stripNulls` properly
we'll need to make use of [Dart's reflection capabilities](https://api.dart.dev/stable/2.19.4/dart-mirrors/dart-mirrors-library.html), the extent of which I haven't fully explored but presumably we'll be able to do some runtime dynamic trickery to figure out whether the property actually exists on `params`. I didn't want to block the login feature at the time but I'll add a separate ticket for this 👍🏻

Widcket (approved):

PR #224, opened 2023-03-15
[SDK-4002] Add support for loginWithPopup

stevehobbsdev:

### 📋 Changes

This adds support for `loginWithPopup` in the client SDK.

Usage:

```dart
await auth0.loginWithPopup(
        audience: 'http://my.api',
        organizationId: 'org123',
        invitationUrl: 'http://invitation.url',
        scopes: {'openid'},
        maxAge: 20,
        popupWindow: window,
        timeoutInSeconds: 120);
```

image

### DIFF
--- a/auth0_flutter/example/lib/example_app.dart
+++ b/auth0_flutter/example/lib/example_app.dart
@@ -54,7 +54,7 @@ class _ExampleAppState extends State<ExampleApp> {
         // return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
         final creds = await auth0Web.loginWithPopup();
 
-        output = creds?.idToken ?? '';
+        output = creds.idToken;
       } else {
         final result = await webAuth.login();
         output = result.idToken;
--- a/auth0_flutter/example/lib/example_app.dart
+++ b/auth0_flutter/example/lib/example_app.dart
@@ -51,8 +51,9 @@ class _ExampleAppState extends State<ExampleApp> {
     try {
       if (kIsWeb) {
         // return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
-        final creds = await auth0Web.loginWithPopup();
+        final creds = await auth0Web.loginWithPopup(audience: 'js-quickstart');
         output = creds?.idToken ?? '';
+        print(creds?.accessToken);
       } else {
         final result = await webAuth.login();
         output = result.idToken;
--- a/auth0_flutter/example/lib/example_app.dart
+++ b/auth0_flutter/example/lib/example_app.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:html';
 
 import 'package:auth0_flutter/auth0_flutter.dart';
 import 'package:auth0_flutter/auth0_flutter_web.dart';
@@ -51,9 +52,9 @@ class _ExampleAppState extends State<ExampleApp> {
     try {
       if (kIsWeb) {
         // return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
-        final creds = await auth0Web.loginWithPopup(audience: 'js-quickstart');
+        final creds = await auth0Web.loginWithPopup();
+
         output = creds?.idToken ?? '';
-        print(creds?.accessToken);
       } else {
         final result = await webAuth.login();
         output = result.idToken;
--- a/auth0_flutter/example/lib/example_app.dart
+++ b/auth0_flutter/example/lib/example_app.dart
@@ -50,11 +50,13 @@ class _ExampleAppState extends State<ExampleApp> {
     // We also handle the message potentially returning null.
     try {
       if (kIsWeb) {
-        return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
+        // return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
+        final creds = await auth0Web.loginWithPopup();
+        output = creds?.idToken ?? '';
+      } else {
+        final result = await webAuth.login();
+        output = result.idToken;
       }
-
-      final result = await webAuth.login();
-      output = result.idToken;
     } catch (e) {
       output = e.toString();
     }
--- a/auth0_flutter/example/lib/example_app.dart
+++ b/auth0_flutter/example/lib/example_app.dart
@@ -48,14 +48,11 @@ class _ExampleAppState extends State<ExampleApp> {
     // We also handle the message potentially returning null.
     try {
       if (kIsWeb) {
-        // return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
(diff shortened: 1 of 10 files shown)


auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -94,6 +94,45 @@ class Auth0Web {
...
+  /// Additonal notes:
+  ///
+  /// * [audience] relates to the API Identifier you want to reference in your access tokens (see [API settings](https://auth0.com/docs/get-started/apis/api-settings)).
+  /// * [scopes] defaults to `openid profile email`. You can
+  /// override these scopes, but `openid` is always requested regardless of
+  /// this setting.
+  /// * If you want to log into a specific organization, provide the
+  /// [organizationId]. Provide [invitationUrl] if a user has been invited
+  /// to join an organization.
+  /// * To provide your own popup window, create it using the `window.open()`
```
Widcket, line 117:
```suggestion
  /// * To provide your own popup window, create it using the `html.window.open()`
  /// method from the dart:html package.
```

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -94,6 +94,45 @@ class Auth0Web {
           scopes: scopes ?? {},
           idTokenValidationConfig: IdTokenValidationConfig(maxAge: maxAge)));
 
+  /// Opens a popup with the /authorize URL using the parameters provided as
```
Widcket, line 97:
```suggestion
  /// Opens a popup with the `/authorize` URL using the parameters provided as
```

auth0_flutter/lib/auth0_flutter_web.dart
Widcket: Would it still work here? Or should we not expose the `popupWindow` param?

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -94,6 +94,45 @@ class Auth0Web {
...
+  /// * To provide your own popup window, create it using the `window.open()`
+  /// HTML API and set [popupWindow] to the result.
+  Future<Credentials> loginWithPopup(
+          {final String? audience,
+          final String? organizationId,
+          final String? invitationUrl,
+          final int? maxAge,
+          final Set<String>? scopes,
+          final dynamic popupWindow,
+          final int? timeoutInSeconds}) =>
```
Widcket, line 126: No way to pass custom params yet, right?

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -94,6 +94,45 @@ class Auth0Web {
...
+  /// the response is successful, results will be valid according to
+  /// their expiration times.
+  ///
+  /// **Note**: This method should be called from an event handler that was
+  /// triggered by user interaction, such as a button click. Otherwise the
+  /// popup will be blocked in most browsers.
+  ///
+  /// Additonal notes:
+  ///
+  /// * [audience] relates to the API Identifier you want to reference in your access tokens (see [API settings](https://auth0.com/docs/get-started/apis/api-settings)).
```
Widcket, line 110: Doesn't this trigger the 80 chars per column limit?

auth0_flutter/lib/auth0_flutter_web.dart
Widcket: And if it does work, should we type the param as `WindowBase` instead of `dynamic`?

auth0_flutter/example/lib/example_app.dart
```diff
@@ -50,11 +48,14 @@ class _ExampleAppState extends State<ExampleApp> {
     // We also handle the message potentially returning null.
     try {
       if (kIsWeb) {
-        return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
-      }
+        // return auth0Web.loginWithRedirect(redirectUrl: 'http://localhost:3000');
+        final creds = await auth0Web.loginWithPopup();
```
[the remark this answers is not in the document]
stevehobbsdev: Definitely, this was an error. Fixed in 9c383dc

auth0_flutter/lib/auth0_flutter_web.dart
stevehobbsdev: Yes it works.

```dart
final popup = window.open('', '', 'width=400,height=800');
final creds = await auth0Web.loginWithPopup(popupWindow: popup);
```

It does however require that `dart:html` is imported. This is why I initially typed it as `dynamic` in the API, but I've also just noticed that the developer would also receive [a warning](https://dart-lang.github.io/linter/lints/avoid_web_libraries_in_flutter.html) in the package plugin that says web libraries should not be used.

So perhaps we _can't_ support it without asking the developer to do some wrangling, for which we should provide an example. I'll leave it to you to decide how we proceed here.

stevehobbsdev: I would say that supporting a custom popup window is sometimes necessary depending on the browser, I think Safari is particular finicky about the popup not being opened directly from user interaction and will block the default one created by the SDK.

auth0_flutter/lib/auth0_flutter_web.dart
stevehobbsdev: We agreed previously that this would be done in a separate ticket & PR.

auth0_flutter/lib/auth0_flutter_web.dart
stevehobbsdev: No, it looks to ignore lines that contain a URL. However, have adjusted it anyway in 0fc7a30

auth0_flutter/lib/auth0_flutter_web.dart
Widcket: Thanks for the clarification, I can see the issue. I'd suggest leaving it as `dynamic`, so it doesn't pull the `dart:html` simply by importing the SDK, but still allowing the developer to provide a custom popup if necessary.
I'd also suggest to document this in the API docs for the method.

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -94,6 +94,60 @@ class Auth0Web {
...
+  /// this setting.
+  /// * If you want to log into a specific organization, provide the
+  /// [organizationId]. Provide [invitationUrl] if a user has been invited
+  /// to join an organization.
+  ///
+  ///
+  /// ### Using a custom popup
+  /// To provide your own popup window, create it using the `window.open()`
+  /// HTML API and set [popupWindow] to the result. You may want to do this
+  /// if certain browsers (e.g. Safari) block the popup by default; in this
```
Widcket, line 123:
```suggestion
  /// if certain browsers (like Safari) block the popup by default; in this
```

Widcket (approved):
LGTM

PR #232, opened 2023-03-31
[SDK-4110] Support options to `getTokenSilently`

stevehobbsdev:

### 📋 Changes

This PR adds support for the options to `getTokenSilently` in the underlying SDK, with the exception of `redirect_uri` (Ionic has a specific use case/need for this that doesn't apply here).

### 🎯 Testing

Covered with unit tests, also tested manually.

### DIFF
--- a/auth0_flutter/lib/src/web/js_interop.dart
+++ b/auth0_flutter/lib/src/web/js_interop.dart
@@ -98,14 +98,11 @@ class Auth0ClientOptions {
 @JS()
 @anonymous
 class GetTokenSilentlyAuthParams {
-  external String? redirect_uri;
   external String? scope;
   external String? audience;
 
   external factory GetTokenSilentlyAuthParams(
-      {final String? redirect_uri,
-      final String? audience,
-      final String? scope});
+      {final String? audience, final String? scope});
 }
 
 @JS()
--- a/auth0_flutter/lib/src/web/js_interop.dart
+++ b/auth0_flutter/lib/src/web/js_interop.dart
@@ -97,17 +97,30 @@ class Auth0ClientOptions {
 
 @JS()
 @anonymous
+class GetTokenSilentlyAuthParams {
+  external String? redirect_uri;
+  external String? scope;
+  external String? audience;
+
+  external factory GetTokenSilentlyAuthParams(
+      {final String? redirect_uri,
+      final String? audience,
+      final String? scope});
+}
+
+@JS()
+@anonymous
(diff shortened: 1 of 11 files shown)


auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -95,19 +95,32 @@ class Auth0ClientOptions {
       final bool? useRefreshTokensFallback});
 }
 
+@JS()
+@anonymous
+class GetTokenSilentlyAuthParams {
+  external String? redirect_uri;
```
Widcket, line 101: Should we specify `redirect_uri` here if we won't use it?

auth0_flutter/lib/src/web/js_interop_utils.dart
```diff
@@ -41,4 +41,10 @@ extension AuthParamsExtension on AuthorizationParams {
   AuthorizationParams prepare([final Map<String, dynamic>? params]) =>
       JsInteropUtils.stripNulls(
           JsInteropUtils.addCustomParams(this, params ?? {}));
+
+  // Converts an instance of AuthorizationParams to
+  // GetTokenSilentlyAuthorizationParams.
+  GetTokenSilentlyAuthParams toGetTokenSilentlyParams() =>
+      JsInteropUtils.stripNulls(GetTokenSilentlyAuthParams(
+          redirect_uri: redirect_uri, scope: scope, audience: audience));
```
Widcket, line 49: Do we want to pass `redirect_uri` here?

auth0_flutter_platform_interface/lib/src/web/cache_mode.dart
```diff
@@ -0,0 +1,17 @@
+enum CacheMode {
```
Widcket, line 1: Should this type have API docs, since it's going to be exposed to (and used by) developers?

auth0_flutter/lib/src/web/js_interop_utils.dart
stevehobbsdev: Nope, removed in 30817f7

auth0_flutter_platform_interface/lib/src/web/cache_mode.dart
stevehobbsdev: Totally! I had actually done them but didn't save the file, so they didn't get included in the commit 🤦🏻 apologies!

Updated in b8e52e4

Widcket (approved):

PR #206, opened 2023-02-24
fix Arabic date parsing issue in Android

mostafaac30:
- [ ] All new/changed/fixed functionality is covered by tests (or N/A)
- [ ] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

I recently submitted a pull request on the Auth0 repository on GitHub regarding an issue with parsing Arabic dates in the auth0_flutter package. The issue was that the package was unable to parse Arabic dates, leading to a FormatException when attempting to convert them to DateTime objects.

To solve this issue, I added a method to the Credentials class that converts Arabic date strings to their English equivalents before parsing them as DateTime objects. This solution allows the package to properly handle Arabic dates, which is particularly important for apps that may be used in Arabic-speaking countries.

### 🎯 Testing

The auth0_flutter_platform_interface version should be updated or you should override dependencies at ` pubspec.yaml` 

```
dependency_overrides:
  auth0_flutter_platform_interface:
    path: ../../auth0_flutter_platform_interface
```
May required:

`publish_to: "none"
`
When using it directly as :

```
dependencies:
  auth0_flutter_platform_interface:
    path: ../auth0_flutter_platform_interface

```

PR #238, opened 2023-04-11
Use Locale.US for Android

Mecharyry:

### 📋 Changes

Offering as an alternative to https://github.com/auth0/auth0-flutter/pull/206.

This aligns how Android and iOS parse dates. iOS is already formatting dates to `ISO8601String` by directly taking the `expiresAt` property, without modification. Android on the other hand is modifying this `expiresAt` by allowing it to be localised to whatever the systems `getDefaultLocale` is. 

To solve the issues we have seen with `ar` this ensures that all dates are using a consistent Locale of `Locale.US`. The test added `credentials_test` ~~is used just to illustrate the breaking change~~ is used to illustrate the issue with Arabic vs US English, but is actually fixed at the native level.

### 📎 References

No references.

### 🎯 Testing

Put your device into `Arabic` and attempt to use this library to log in. You'll notice that it works in this MR but doesn't on `main`

### DIFF
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
@@ -55,7 +55,7 @@ class LoginApiRequestHandler : ApiRequestHandler {
             override fun onSuccess(credentials: Credentials) {
                 val scope = credentials.scope?.split(" ") ?: listOf()
                 val sdf =
-                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.US)
+                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US)
 
                 val formattedDate = sdf.format(credentials.expiresAt)
                 result.success(
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
@@ -55,7 +55,7 @@ class LoginApiRequestHandler : ApiRequestHandler {
             override fun onSuccess(credentials: Credentials) {
                 val scope = credentials.scope?.split(" ") ?: listOf()
                 val sdf =
-                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.getDefault())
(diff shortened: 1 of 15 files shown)


Mecharyry:
> Thankss @Mecharyry - can you elaborate on the breaking change? Does it introduce one?

Sorry I goofed when writing the description. I meant to write that the test class illustrates why parsing an Arabic specific date time breaks vs a US one. I've updated the description.

poovamraj:
@stevehobbsdev I'd prefer this solution as well over the other. LGTM 👍

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginApiRequestHandler.kt
```diff
@@ -55,7 +55,7 @@ class LoginApiRequestHandler : ApiRequestHandler {
             override fun onSuccess(credentials: Credentials) {
                 val scope = credentials.scope?.split(" ") ?: listOf()
                 val sdf =
-                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.getDefault())
+                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.US)
```
poovamraj, line 58: Can we use the format `yyyy-MM-dd'T'HH:mm:ss.SSS'Z'` as that is the one used in Android and iOS.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/LoginWithOtpApiRequestHandler.kt
```diff
@@ -43,7 +43,7 @@ class LoginWithOtpApiRequestHandler: ApiRequestHandler {
             override fun onSuccess(credentials: Credentials) {
                 val scope = credentials.scope?.split(" ") ?: listOf()
                 val sdf =
-                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.getDefault())
+                    SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ", Locale.US)
```
poovamraj, line 46: Same as above for all the format strings

poovamraj (changes requested):

Mecharyry:
@poovamraj Sorry for the delay, I've made the requested changes!

Mecharyry:
@poovamraj @stevehobbsdev Can I get some eyes on this? It would be great to merge this so I can stop maintaining a fork 😄

stevehobbsdev:
@Mecharyry Thanks for your patience here. We're gearing up for a new release that includes web support, this PR will be included along with that work. We're aiming for that release within the next week or so.

stevehobbsdev (approved):

poovamraj (approved):
Checked the repo for use of Locale.getDefault(). Looks fine to me +1

PR #247, opened 2023-04-14
af: Bump `auth0_flutter_platform_interface` to beta version

Widcket:

### 📋 Changes

This PR bumps the version of `auth0_flutter_platform_interface` to [v1.2.0-beta.0](https://pub.dev/packages/auth0_flutter_platform_interface/versions/1.2.0-beta.0).

### DIFF
--- a/auth0_flutter/example/pubspec.yaml
+++ b/auth0_flutter/example/pubspec.yaml
@@ -31,10 +31,6 @@ dependencies:
     sdk: flutter
   flutter_dotenv: ^5.0.2
 
-dependency_overrides:
-  auth0_flutter_platform_interface:
-    path: ../../auth0_flutter_platform_interface
-
 dev_dependencies:
   # The "flutter_lints" package below contains a set of recommended lints to
   # encourage good coding practices. The lint set provided by the package is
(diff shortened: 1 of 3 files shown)


auth0_flutter/example/pubspec.yaml
```diff
@@ -31,10 +31,6 @@ dependencies:
     sdk: flutter
   flutter_dotenv: ^5.0.2
 
-dependency_overrides:
-  auth0_flutter_platform_interface:
-    path: ../../auth0_flutter_platform_interface
-
```
stevehobbsdev, line 37: I wonder if we should leave this in place for ease of development, as it's a pain to continually have to adjust the spec for local development vs deployed.

It [looks like this node is ignored anyway for consumers of published packages](https://dart.dev/tools/pub/dependencies#:~:text=Only%20the%20dependency,of%20your%20package.):

> Only the dependency overrides in a package’s own pubspec are considered during package resolution. Dependency overrides inside any depended-on packages are ignored.
>
> As a result, if you publish a package to pub.dev, keep in mind that your package’s dependency overrides are ignored by all users of your package.

Widcket: Good catch! Restored in https://github.com/auth0/auth0-flutter/pull/247/commits/00e149f30440a01dd5342a3a746ccd9dd7740dc8

stevehobbsdev (approved):

PR #282, opened 2023-07-03
[SDK-3508] Fix Integration Test CI for Android

poovamraj:

### 📋 Changes

We have fixed the CI for Android Integration test in this PR. 
- We have used appium to run these tests as it is more reliable. 
- The matcher to find login in Android Integration Test is using `hints` that has been provided Appium Inspector as resource-id is not present in classic UL where as new UL has them.
- We also screen record the integration test in Android
- We use xlarge resource class as large and below causes emulator to fail at times.

### 📎 References
http://appium.io/docs/en/2.0/
https://discuss.circleci.com/t/failed-to-run-python-script-directly-pyenv-version-name-command-not-found/37526/2
https://github.com/CircleCI-Public/flutter-orb/blob/main/src/scripts/install-sdk.sh#L21

### 🎯 Testing

This PR adds integration tests


.circleci/config.yml
```diff
@@ -1,10 +1,58 @@
 version: 2.1
 orbs:
-  android: circleci/android@2
+  android: circleci/android@2.3.0
```
Widcket, line 3: Should we pin it to the v2 major?

.circleci/config.yml
```diff
@@ -5,6 +5,54 @@ orbs:
   flutter: circleci/flutter@2
   browser-tools: circleci/browser-tools@1
   codecov: codecov/codecov@3
+  node: circleci/node@5.1.0
```
Widcket, line 8: Let's also pin this orb to the major version.

.circleci/config.yml
```diff
@@ -5,6 +5,54 @@ orbs:
...
+  flutter-install:
+    orbs:
+      android: circleci/flutter@2
+    commands:
+      install_sdk:
+        description: Custom script to fix issue in Flutter orb
```
Widcket, line 14: Which issue is this fixing?

poovamraj: This is mentioned in the PR description - There was also another [issue](https://discuss.circleci.com/t/failed-to-run-python-script-directly-pyenv-version-name-command-not-found/37526/2) with the CI where the [Flutter Orb](https://github.com/CircleCI-Public/flutter-orb/blob/main/src/scripts/install-sdk.sh#L21) which was exporting the flutter binary in a way that affected. To fix this we have copied the bash script and patched the required changes

.circleci/config.yml
```diff
@@ -5,6 +5,54 @@ orbs:
...
+        parameters:
+          version:
+            type: string
+            description: The target version for the Flutter SDK.
+            default: "3.10.2"
```
Widcket, line 19: Let's use the latest Flutter version.

.circleci/config.yml
Widcket: The linked forum thread is about a Python script. Could you please expand about the issue in the Flutter Orb?

poovamraj: You can see in the next link that the flutter-orb is also making the same mistake as mentioned. The flutter-orb is adding the flutter binary to the path in a way which affects other paths. This link mentioned in the forum explains it better - https://github.com/CircleCI-Public/gcp-cli-orb/issues/22#issuecomment-715463423

Widcket: Thanks for the context. Wasn't that fixed here https://github.com/CircleCI-Public/flutter-orb/pull/28/files#diff-8f89d2af8215d6241d3412ffb02561f587288984cf9de7cacb93e5df1177e81eR21?

Widcket: (They escaped the `$` in `$PATH`)

appium-test/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "auth0-flutter-test",
+  "version": "1.0.0",
+  "description": "Integration Test for Auth0 Flutter",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
```
Widcket, line 10:
```suggestion
  "license": "Apache-2.0",
```
The SDK is using the Apache v2 license.

appium-test/test.js
```diff
@@ -0,0 +1,45 @@
+const {remote} = require('webdriverio');
+
+const capabilities = {
+  platformName: 'Android',
+  'appium:automationName': 'UiAutomator2',
+  'appium:deviceName': 'Android',
+  'appium:appPackage': 'com.auth0.auth0_flutter_example',
+  'appium:appActivity': '.MainActivity',
+  'appium:app': '/home/circleci/project/auth0_flutter/example/build/app/outputs/flutter-apk/app-x86_64-release.apk'
```
Widcket, line 9: Any way we can avoid having an absolute path here?

Widcket (changes requested):
Thanks so much for fixing this!

Just a couple things:
- Let's stick, if possible, to the Circle CI Flutter Orb unless absolutely necessary.
- Let's try to avoid hardcoded absolute paths whenever possible.

.circleci/config.yml
poovamraj: Great catch, looks like I was using older version of the orb for testing. This is fixed now and the description is updated

appium-test/test.js
poovamraj: Good catch, this is now updated

poovamraj: Good catch, this is now updated

poovamraj:
Hey @Widcket, Great review. I have updated both the concerns and they should GTG now.

.circleci/config.yml
```diff
@@ -133,33 +110,86 @@ jobs:
...
+      image: android:2022.12.1
+      # To optimize build times, we recommend "large" and above for Android-related jobs
+      resource_class: xlarge
     steps:
-      - run:
-          name: Guard Smoke Tests
-          command: |
-            if ! [[ -z "$CIRCLE_PR_NUMBER" ]]; then
-                circleci-agent step halt
-            fi
```
Widcket, line 149: One last thing, we should probably keep this to avoid running integration tests on PRs from forks.

Widcket (approved):
LGTM

Issue #233, opened 2023-04-03
Unable to use auth0 in Isolate


eliasjtg:
### Checklist

- [X] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I'm working in a isolate with background location, sending location to backend authenticated with Auth0.

`auth0.credentialsManager.credentials()`

This show the following exception:

```
E/MethodChannel#auth0.com/auth0_flutter/credentials_manager( 6251): Failed to handle method call
E/MethodChannel#auth0.com/auth0_flutter/credentials_manager( 6251): kotlin.UninitializedPropertyAccessException: lateinit property activity has not been initialized
E/MethodChannel#auth0.com/auth0_flutter/credentials_manager( 6251): 	at com.auth0.auth0_flutter.CredentialsManagerMethodCallHandler.getActivity(CredentialsManagerMethodCallHandler.kt:25)
E/MethodChannel#auth0.com/auth0_flutter/credentials_manager( 6251): 	at com.auth0.auth0_flutter.CredentialsManagerMethodCallHandler.onMethodCall(CredentialsManagerMethodCallHandler.kt:36)
E/MethodChannel#auth0.com/auth0_flutter/credentials_manager( 6251): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:258)
E/MethodChannel#auth0.com/auth0_flutter/credentials_manager( 6251): 	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:295)
[...]

PR #289, opened 2023-07-11
af: Use Application Context for Credentials Manager

poovamraj:
- [ ] I have added documentation for all new/changed functionality - N/A - Changes are internal

### 📋 Changes
We are using `applicationContext` to fetch the credentials instead of `Activity` which is not required and could cause issues when credentials are fetched from outside an activity

### 📎 References
https://github.com/auth0/auth0-flutter/issues/233

### 🎯 Testing

We have updated the existing Unit Tests and added check that proper context is set.

### DIFF
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsManagerMethodCallHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsManagerMethodCallHandler.kt
@@ -1,6 +1,7 @@
 package com.auth0.auth0_flutter
 
 import android.app.Activity
+import android.content.Context
 import android.content.Intent
 import androidx.annotation.NonNull
 import com.auth0.android.authentication.AuthenticationAPIClient
@@ -23,6 +24,8 @@ class CredentialsManagerMethodCallHandler(private val requestHandlers: List<Cred
     // GitHub issue: https://github.com/jacoco/jacoco/issues/1182
     @Generated
     lateinit var activity: Activity
+    @Generated
+    lateinit var context: Context
 
     var credentialsManager: SecureCredentialsManager? = null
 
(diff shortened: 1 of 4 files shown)


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsManagerMethodCallHandler.kt
```diff
@@ -23,6 +24,8 @@ class CredentialsManagerMethodCallHandler(private val requestHandlers: List<Cred
     // GitHub issue: https://github.com/jacoco/jacoco/issues/1182
     @Generated
     lateinit var activity: Activity
```
Widcket, line 26: Do we still need this?

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/CredentialsManagerMethodCallHandlerTest.kt
```diff
@@ -33,12 +34,14 @@ class CredentialsManagerMethodCallHandlerTest {
         arguments: HashMap<String, Any?> = defaultArguments,
         requestHandlers: List<CredentialsManagerRequestHandler>,
         activity: Activity? = null,
```
Widcket, line 36: Is this still needed?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsManagerMethodCallHandler.kt
poovamraj: Yes, as you can see `requireAuthentication` still needs `activity`

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/CredentialsManagerMethodCallHandlerTest.kt
poovamraj: As mentioned in the previous comment, we still use it in `requireAuthentication` so it is required in tests as well

Widcket (approved):

Issue #286, opened 2023-07-09
[Android] expiration date mismatch


o-artebiakin:
### Checklist

- [X] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Credentials Manager for the Android app is returning an invalid expiration time. The credentialsManager.credentials() method is returning a Credentials object, but the expiresAt field does not match the expiration date provided in the idToken. This mismatch is causing the token to be invalidated on the server.

This issue does not occur on iOS, as it is working correctly.

```
Credentials.expiresAt         =>  2023-07-09 15:15:29.779Z
Credentials.idToken['exp'] => 2023-07-09 21:15:29.000Z
```

### Reproduction

1. Log in to the app.
2. Call `Auth0().credentialsManager.credentials()`.
3. Compare the `expiresAt` field with the expiration date parsed from the `idToken`.

### auth0_flutter version

1.2.1

### Flutter version

3.10.4

### Platform

Android

Widcket:
Hi @o-artebiakin, thanks for raising this.

@poovamraj could you please take a look?

poovamraj:
@o-artebiakin The `expiresAt` field will not match the expiration of `idToken`. The `expiresAt` refers to the expiration of the `access_token` and you should only use that for your APIs. You can find more information on difference between a ID Token and an access token here - https://auth0.com/blog/id-token-access-token-what-is-the-difference/#What-Is-an-ID-Token.

[...]

Issue #160, opened 2022-09-07
credentialsManager alters access token expiry upon storing credentials


tom-smith-alex:
### Checklist

- [X] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [X] I have looked into the [README](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Authenticating with the [API](https://pub.dev/packages/auth0_flutter#api). 
My local timezone is UTC + 10 (Australia/Brisbane). 

**Initial value of expiresAt**
The Credentials returned by `auth0.api.login()` print the following value for the `expiresAt` property (which is defined as "The absolute date and time of when the access token expires"): **2022-09-07 11:04:54.000Z** (or 1662548694000 ms since Epoch). This is expected and in line with the access token expiry value I set in Auth0's dashboard.

**Value of expiresAt after storing credentials**
After invoking the Credentials Manager's `storeCredentials` method to store the above, this is the value of `expiresAt` when retrieved by the `credentials()` method : **2022-09-07 01:04:54.000Z** (or 1662512694000 ms since Epoch).

**Expected**
`expiresAt` should be identical in both scenarios 

**Actual**
Credentials Manager (CM) wrongly adjusts the initial timestamp to my timezone upon storing and / or subsequently retrieving.

**Consequences / side effects**

- `expiresAt`, once stored, cannot be trusted if relied upon (to e.g. have a session timeout warning mechanism when user is inactive)
- upon subsequently calling the CM's `credentials()` method, the wrong timestamp is interpreted and the following message is outputted: "Credentials have expired. Renewing them now..." even though the current credentials should still be valid.
- even when the above happens and the CM wrongly thinks that credentials have expired, its `hasValidCredentials()` method still returns `true`

### Reproduction

[...]

PR #315, opened 2023-10-12
Fix access token expiry android

poovamraj:

### 📋 Changes

This error is happening because in Java we were parsing the credentials result as **Z** which refers to [timezone](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) parsing in Java

But this was later changed to escape as shown below.

Screenshot 2023-10-12 at 13 38 20

This means we were initially parsing the timestamp as UTC at the dart layer when credentials were retrieved and to be same as that we had to set timezone while saving the credentials.

But after that we have removed parsing timezone by escaping the **Z** with quotes `''`. But we haven't removed the timezone being set in save credentials handler. This caused the time to jump to the offset of UTC.

We have fixed this by removing the timezone in save credentials handler.

### 📎 References
https://github.com/auth0/auth0-flutter/issues/286
https://github.com/auth0/auth0-flutter/pull/162
https://github.com/auth0/auth0-flutter/compare/v1.0.1...main

### 🎯 Testing
Existing unit test has been fixed to avoid this issue


Widcket:
Wasn't the Z escaped in this PR: https://github.com/auth0/auth0-flutter/pull/238?

aprzedecki (approved):
Thats a fix we are looking for 👍

Widcket:
In fact, we requested the escaped format:

Screenshot 2023-10-12 at 1 07 04 PM

Are we sure this change won't re-introduce this issue? https://github.com/auth0/auth0-flutter/issues/160

poovamraj:
Yes, escaping it would be the right thing. But while we escaped the timezone "Z" which will parse the date to local time, we have forgotten to remove 

```
format.timeZone = TimeZone.getTimeZone("UTC")
```

Also this would have caused issues for API login and other places since it was set only here and not in the other places where we format date

Widcket:
It'd be great to have a single date formatter (or format string) in the entire Android native layer, to avoid these kinds of issues.

joymyr:
This fixes the token refesh issue for me. Please merge it

Widcket (approved):

aprzedecki:
Great! Can we release it also on pub.dev ?

mng-prophero:
> Great! Can we release it also on pub.dev ?

We're needing this released as well as it's affecting our users

Widcket:
👋🏼 I expect to cut a release by EOW.

PR #346, opened 2023-11-17
af: Gate native Swift/Objective-C code per platform [SDK-4700] 1/3

Widcket:

### 📋 Changes

This PR gates all the iOS-specific code and macOS-specific code in `auth0_flutter` behind compile-time conditionals, so it compiles correctly on both platforms.

Also the Podspec of `auth0_flutter` was updated to use the `FlutterMacOS` dependency on macOS, instead of `Flutter`, which is iOS-only, and the minimum deployment target of the sample app was updated to match the version specified in the Podspec.


Widcket:
The `flutter analyze` CI check (**CI / Analyze auth0_flutter_platform_interface Flutter package**) will pass once https://github.com/auth0/auth0-flutter/pull/347 gets merged.

frederikprijck (approved):

PR #364, opened 2023-12-06
Fix member ordering to fix CI

frederikprijck:


frederikprijck:
codecov is complaining, but I believe we want to ignore that as I dont think we can do much about that.

Widcket (approved):

PR #415, opened 2024-03-05
Avoid Z as it represents UTC

poovamraj:

### 📋 Changes

We are removing using 'Z' in timestamp as it denotes UTC and we are representing the value in local time

### 📎 References
https://github.com/auth0/auth0-flutter/pull/315


Widcket:
Won't this bring back https://github.com/auth0/auth0-flutter/issues/160?

poovamraj:
If you run the example now we get the right values in UTC

```
I/flutter (15545): current time 2024-03-05T17:30:41.365925Z
I/flutter (15545): expiry 2024-03-05 17:35:41.198Z
```

Just add these lines to `webAuthLogin()`

```
      print('current time ${DateTime.now().toUtc().toIso8601String()}');
      print('expiry ${result.expiresAt}');
```

poovamraj:
Issue with #160 was how we formatted data in 1 place and not in the other one. This fix is only on how we show response.

Widcket:
The `result.expiresAt` should be provided in UTC. The consumer code can convert it to local timezone using `toLocal()` if needed.

BTW, this PR is inconsistent with how the time is handled on iOS (also UTC), and also makes the changes in only a couple of Android files (as opposed to all the files that have the format pattern).

PR #468, opened 2024-07-23
Make sure SimpleDateFormat is set to UTC timezone

gferon:

### 📋 Changes

After spending quite some time trying to understand why our access tokens would not be refreshed, I realized that something was wrong with the `expiresAt` field. The issue affects (at least) Android because the UTC timezone is added in the format while the formatter itself is timezone-unaware.

With the following added print statements, you can observe that the `Credentials.expiresAt` is formatted from local time with an appended Z, which means we lose the timezone information.

```
I/System.out(18918): RenewApiRequestHandler::onSuccess credentials.expiresAt Tue Jul 23 08:49:05 GMT+02:00 2024
I/System.out(18918): RenewApiRequestHandler::onSuccess formattedDate 2024-07-23T08:49:05.555Z
```

Making sure all `SimpleDateFormat` are set to the UTC timezone fixes the issue, but since `java.util.Date` is notably confusing to work with, a better fix would maybe be to use `java.time.Instant` (see alternative PR #469) and either call `.toInstant().toString()` everywhere (and `Instant.parse()`) which always work with ISO-8601 representation.

### 🎯 Testing

Print the `expiresAt` date in Flutter with and without this patch. This unit tests didn't catch this problem because the test had the same implementation issue.


desusai7:
Hi @gferon, 

We appreciate your time on raising this PR, however I am closing it as we are considering the PR #469 raised by you !

Thanks

PR #485, opened 2024-11-19
add support for app state argument

navaronbracke:
Fixes https://github.com/auth0/auth0-flutter/issues/486

### 📋 Changes

This PR adds support for the `appState` parameter to `loginWithRedirect()` for the web.
See https://github.com/auth0/auth0-spa-js/blob/f2e566849efa398ca599daf9ebdfbbd62fcb1894/src/global.ts#L298

There is prior art in https://github.com/auth0/auth0-angular/pull/168

### 🎯 Testing

Added several unit tests. This can be tested end-to-end by setting up a login with redirect, that uses the app state argument as parameter, and then validating that the app state is returned when the user is redirected back.
I did verify end-to-end using an internal application that uses the SDK integration.

To run the unit tests locally, use `flutter test test/web/auth0_flutter_web_test.dart --platform=chrome`


auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -160,6 +161,32 @@ void main() {
     expect(params.max_age, null);
   });
 
+  test('loginWithRedirect supports appState parameter', () async {
```
navaronbracke, line 164: I ran the tests using `flutter test --platform=chrome --tags=browser -v` but I didn't see my test name in the verbose output? Is there a way to run this test locally?

auth0_flutter/pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependency_overrides:
 
 dev_dependencies:
   build_runner: ^2.1.8
+  collection: ^1.18.0
```
navaronbracke, line 24: For the `MapEquality` in the test, hence a dev dependency

navaronbracke:
It seems I still missed a step, as the app state is not passed to the redirect url when actually coming back to the app. Once that is fixed, this is ready for review.

auth0_flutter/lib/src/web/auth0_flutter_web_platform_proxy.dart
```diff
@@ -20,8 +20,8 @@ class Auth0FlutterWebClientProxy {
           [final GetTokenSilentlyOptions? options]) =>
       promiseToFuture(client.getTokenSilently(options));
 
-  Future<void> handleRedirectCallback() =>
-      promiseToFuture(client.handleRedirectCallback());
+  Future<RedirectLoginResult> handleRedirectCallback([final String? url]) =>
```
navaronbracke, line 23: The return value and parameters didn't match what is defined in the JS SDK.

For the fix, I need the new return value, but the url is currently unused.
Adding support for custom URL's could be done later, I guess.

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -177,12 +191,17 @@ class PopupConfigOptions {
...
   external Auth0Client(final Auth0ClientOptions options);
   external Future<void> loginWithRedirect([final RedirectLoginOptions options]);
-  external Future<void> loginWithPopup(
-      [final PopupLoginOptions? options, final PopupConfigOptions? config]);
-  external Future<void> handleRedirectCallback([final String? url]);
+  external Future<void> loginWithPopup([
+    final PopupLoginOptions? options,
+    final PopupConfigOptions? config,
+  ]);
+  external Future<RedirectLoginResult> handleRedirectCallback([
```
navaronbracke, line 198: The app state is defined in the RedirectLoginResult, so we need it here too.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -23,20 +23,54 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
...
+  /// The app state that was passed through [loginWithRedirect]
+  /// and retrieved in [initialize].
+  ///
+  /// This object is always a Dart object, never a JS object.
+  ///
+  /// When the login completes with the redirect, the page is reloaded.
+  /// Thus clearing this object is not needed,
+  /// as the actual state is managed across reloads,
+  /// using the transaction manager.
+  Object? _appState;
```
navaronbracke, line 35: This property & getter are new. We provide these now, so that:
a) users can retrieve the app state
b) we do not need to change the API for initialize/onLoad (the app state is web only anyway? Or so I think?)

auth0_flutter/lib/src/web/js_interop_utils.dart
```diff
@@ -31,4 +31,22 @@ class JsInteropUtils {
 
     return obj;
   }
+
+  // TODO: replace with `dartify` from `dart:js_interop_unsafe` when migrating to WASM
```
navaronbracke, line 35: I added these here, to make the migration to WASM easier. Then we only need to change these two, instead of all callsites

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -72,13 +74,75 @@ void main() {
 
   test('handleRedirectCallback is called on load when auth params exist in URL',
       () async {
+    final interop.RedirectLoginResult mockRedirectResult =
```
navaronbracke, line 77: This test was updated. Now it also mocks `handleRedirectCallback`

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -72,13 +74,75 @@ void main() {
...
+    when(mockClientProxy.handleRedirectCallback())
+        .thenAnswer((final _) => Future.value(mockRedirectResult));
+
+    plugin.urlSearchProvider = () => '?code=abc&state=123';
+    await auth0.onLoad();
+    verify(mockClientProxy.handleRedirectCallback());
+    verifyNever(mockClientProxy.checkSession());
+  });
+
+  test('handleRedirectCallback captures appState that was passed', () async {
```
navaronbracke, line 91: New test to verify behavior of the getter

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -72,13 +74,75 @@ void main() {
...
+
+    expect(capturedAppState, isNotNull);
+    expect(capturedAppState, isA<Map<Object?, Object?>>());
+    capturedAppState as Map<Object?, Object?>;
+    const MapEquality<Object?, Object?> eq = MapEquality<Object?, Object?>();
+
+    expect(eq.equals(capturedAppState, appState), isTrue);
+  });
+
+  test('appState getter returns null when accessed more than once', () async {
```
navaronbracke, line 121: New test to verify behavior of the app state

navaronbracke:
Failing test was fixed, so this is ready for review!

navaronbracke:
@Widcket Hi! I wanted to ask if you have any updates regarding a code review for this change? Thanks!

Widcket:
Hi @navaronbracke, first we need to sort out some CI issues. Thank you for the contribution, though. We haven't forgotten about it and will be able to get to it once we are able to.

navaronbracke:
Thanks! And I'll address the formatting in this PR. I see that one of the tasks is now failing on the line length in one of the added files.

auth0_flutter/lib/src/web/js_interop_utils.dart
```diff
@@ -22,7 +22,9 @@ class JsInteropUtils {
   // **Note**: there is no static typing for these parameters to be able
   // to retrieve them again.
   static T addCustomParams<T extends Object>(
-      final T obj, final Map<String, dynamic> params) {
+    final T obj,
+    final Map<String, dynamic> params,
```
navaronbracke, line 26: Adding a comma here was needed for the formatter to properly format this under 80 characters per line.

auth0_flutter/lib/src/web/js_interop_utils.dart
```diff
@@ -31,4 +33,22 @@ class JsInteropUtils {
 
     return obj;
   }
+
+  // TODO: replace w/ `dartify` from `dart:js_interop_unsafe` when migrating to WASM
```
navaronbracke, line 37: Apparently comments are also checked for line length, which is a bit odd.

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -72,13 +74,91 @@ void main() {
...
+
+    expect(capturedAppState, isNotNull);
+    expect(capturedAppState, isA<Map<Object?, Object?>>());
+    capturedAppState as Map<Object?, Object?>;
+    const MapEquality<Object?, Object?> eq = MapEquality<Object?, Object?>();
+
+    expect(eq.equals(capturedAppState, appState), isTrue);
+  });
+
+  test('appState getter throws UnimplementedError when not overridden',
```
navaronbracke, line 121: This test was added as well. While most of the coverage report was for files that I didn't change, the report did notify about two untested lines in the default implementation of the getter.

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -19,6 +19,15 @@ class Auth0Web {
...
       : _account = Account(domain, clientId);
 
+  /// Get the app state that was provided during a previous call
+  /// to [loginWithRedirect].
+  ///
+  /// This method should be called after calling [onLoad].
+  ///
+  /// This getter can only be called once,
+  /// after which it will return `null` for subsequent invocations.
+  Future<Object?> get appState => Auth0FlutterWebPlatform.instance.appState;
```
Widcket, line 29: What is the purpose of having `appState` return `null` after the first invocation? I don't think the `appState` from the SPA JS behaves this way...

navaronbracke: I can change that, sure. We should be porting the JS API 1-1, so this should match indeed.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -23,20 +23,54 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
...
+
+  @override
+  Future<Object?> get appState {
+    if (_appState != null) {
+      final Object? appState = _appState;
```
Widcket, line 40: Should this be of type `Object` instead of `Object?`?

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
Widcket: Good point on avoiding the braking change. Would you mind noting this down in a comment?

auth0_flutter/test/web/auth0_flutter_web_test.dart
Widcket: I guess this test should be removed, since `appState` will no longer be `null` after the first access.

auth0_flutter/test/web/auth0_flutter_web_test.dart
Widcket: It's still reporting these lines as uncovered
Screenshot 2025-02-10 at 21 57 43

Widcket:
Thanks for your contribution, it's pretty thorough. I've left a few comments, once these get addressed and the Codecov checks pass, I'll test manually before approving.

auth0_flutter/test/web/auth0_flutter_web_test.dart
navaronbracke: I'll modify the test instead, to verify that app state is the same when accessed multiple times

auth0_flutter/test/web/auth0_flutter_web_test.dart
navaronbracke: Hmm, I'll have a look at it!

Widcket:
Also, the commits must be verified before merging.

navaronbracke:
Oh, okay. Can I use a rebase to sign the unsigned commits in this PR? Or is that against your terms of service / code of conduct?

Widcket:
Yes, a rebase is fine.

Issue #463, opened 2024-06-11
Add Support for WASM Compilation in Flutter 3.22


EduardoYamauchi:
### Checklist

- [X] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Hello,

With the release of Flutter 3.22, the WASM (WebAssembly) compiler can now be integrated. I believe it has become the standard for Web applications at some point.

https://docs.flutter.dev/platform-integration/web/wasm

Please add support for compiling with WASM. `flutter build web --release --wasm`.

Thank you!

### Reproduction

`flutter build web --release --wasm`

### auth0_flutter version

1.7.2

### Flutter version

3.22

### Platform

Web

slavap:
This package needs to be switched from package:js to package:web and package:js_interop

kod3000:
same here. WASM is now stable in flutter please make this a priority.

MrObe13:
Any news about that?

kod3000:
I attempted to rewrite this myself cause as mention its a thing of changing out old dependencies, but encountered some challenges. It seems that Flutter support is minimal as they have a backlog going to Mar 2023. Looking at the ppl who are actually committing it appears to be outsourced rather than handled internally. I'm seriously considering moving away from this platform entirely, as their lack of active development is concerning—especially in the security sector, where consistent updates are crucial.

Nitrono:
Any news on it? This is the only package which stops our project from moving to wasm

[...]

PR #516, opened 2025-02-18
Added support for WASM compilation

pmathew92:
### 📋 Changes

This PR adds support for WASM compilation 

### 📎 References
#463 

### 🎯 Testing
This has been tested by running the example app with the  following command
`flutter run -d chrome --web-port 3000 --wasm `

### DIFF
--- a/auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
+++ b/auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
@@ -10,11 +10,11 @@ extension WebExceptionExtension on WebException {
     final error = jsException.toJSBox.getProperty<JSString>('error'.toJS);
     final description =
         jsException.toJSBox.getProperty<JSString>('error_description'.toJS);
-    final Map<String, JSAny> details = {};
+    final Map<String, JSAny?> details = {};
 
     keys(jsException.toJSBox).toDart.forEach((final key) {
-      if (key == 'error' || key == 'error_description') return;
-      details[key.toString()] = jsException.toJSBox.getProperty<JSAny>(key);
+      if (key == 'error'.toJS || key.toDart == 'error_description'.toJS) return;
+      details[key.toDart] = jsException.toJSBox.getProperty<JSAny?>(key);
     });
 
     switch (error) {
@@ -31,10 +31,10 @@ extension WebExceptionExtension on WebException {
       case 'unsupported_response_type':
       case 'unsupported_grant_type':
       case 'temporarily_unavailable':
-        return WebException.authenticationError(error.toString(),
-            description.toString(), {'state': details['state']});
+        return WebException.authenticationError(error.toDart,
+            description.toDart, {'state': details['state']});
       case 'mfa_required':
-        return WebException.mfaError(description.toString(),
+        return WebException.mfaError(description.toDart,
             jsException.toJSBox.getProperty('mfaToken'.toJS));
       case 'timeout':
         return WebException.timeout(description.toDart);
@@ -44,6 +44,6 @@ extension WebExceptionExtension on WebException {
         return WebException.missingRefreshToken(description.toDart);
     }
 
-    return WebException(error.toString(), description.toDart, details);
+    return WebException(error.toDart, description.toDart, details);
   }
 }
--- a/auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
+++ b/auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
@@ -6,15 +6,15 @@ import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interfac
 external JSArray<JSString> keys(final JSObject o);
 
 extension WebExceptionExtension on WebException {
-  static WebException fromJsObject(final Object jsException) {
-    final error = jsException.toJSBox.getProperty<JSString>('error'.toJS);
+  static WebException fromJsObject(final JSObject jsException) {
+    final error = jsException.getProperty<JSString>('error'.toJS);
     final description =
-        jsException.toJSBox.getProperty<JSString>('error_description'.toJS);
+        jsException.getProperty<JSString>('error_description'.toJS);
     final Map<String, JSAny?> details = {};
 
-    keys(jsException.toJSBox).toDart.forEach((final key) {
-      if (key == 'error'.toJS || key.toDart == 'error_description'.toJS) return;
-      details[key.toDart] = jsException.toJSBox.getProperty<JSAny?>(key);
+    keys(jsException).toDart.forEach((final JSString key) {
+      if (key.toDart == 'error' || key.toDart == 'error_description') return;
+      details[key.toDart] = jsException.getProperty<JSAny?>(key);
     });
 
     switch (error) {
@@ -31,11 +31,11 @@ extension WebExceptionExtension on WebException {
       case 'unsupported_response_type':
       case 'unsupported_grant_type':
       case 'temporarily_unavailable':
-        return WebException.authenticationError(error.toDart,
-            description.toDart, {'state': details['state']});
+        return WebException.authenticationError(
+            error.toDart, description.toDart, {'state': details['state']});
       case 'mfa_required':
         return WebException.mfaError(description.toDart,
-            jsException.toJSBox.getProperty('mfaToken'.toJS));
+            jsException.getProperty('mfaToken'.toJS) as String);
       case 'timeout':
         return WebException.timeout(description.toDart);
       case 'cancelled':
--- a/auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
+++ b/auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
@@ -1,15 +1,20 @@
-import 'dart:js_util';
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
 import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interface.dart';
+
+@JS('Object.keys')
+external JSArray<JSString> keys(final JSObject o);
 
 extension WebExceptionExtension on WebException {
   static WebException fromJsObject(final Object jsException) {
-    final error = getProperty<String>(jsException, 'error');
-    final description = getProperty<String>(jsException, 'error_description');
-    final Map<String, dynamic> details = {};
+    final error = jsException.toJSBox.getProperty<JSString>('error'.toJS);
+    final description =
+        jsException.toJSBox.getProperty<JSString>('error_description'.toJS);
+    final Map<String, JSAny> details = {};
 
-    objectKeys(jsException).forEach((final key) {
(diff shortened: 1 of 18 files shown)


auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -52,29 +53,29 @@ class RedirectLoginOptions {
...
 
 @JS()
 @anonymous
-class Auth0ClientInfo {
-  external Map<String, String>? get env;
-  external String get name;
-  external String get version;
+extension type Auth0ClientInfo._(JSObject _) implements JSObject {
+  external JSObject?  env;
+  external String  name;
```
navaronbracke, line 67: The `get` is gone here?

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -96,7 +97,7 @@ class Auth0ClientOptions {
 
 @JS()
 @anonymous
-class GetTokenSilentlyAuthParams {
+extension type GetTokenSilentlyAuthParams._(JSObject _ )implements JSObject {
```
navaronbracke, line 100: Missing space after the )

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -165,24 +166,25 @@ class PopupLoginOptions {
 
 @JS()
 @anonymous
-class PopupConfigOptions {
-  external dynamic get popup;
-  external int? get timeoutInSeconds;
+extension type PopupConfigOptions._ (JSObject _ ) implements JSObject  {
+  external JSAny  popup;
```
navaronbracke, line 170: This should be `JSAny?` and there is a space too many. The `get` is also gone here?

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -165,24 +166,25 @@ class PopupLoginOptions {
...
+  external int?  timeoutInSeconds;
 
   external factory PopupConfigOptions(
-      {final dynamic popup, final int? timeoutInSeconds});
+      {final JSAny popup, final int? timeoutInSeconds});
```
navaronbracke, line 174: Shouldn't this be `JSAny?` for the popup argument?

navaronbracke (changes requested):
Just a quick review, as I have experience with WASM interop from other projects.
(and this might also affect one of my changes to this plugin, hence the added interest)

pmathew92:
> Just a quick review, as I have experience with WASM interop from other projects. (and this might also affect one of my changes to this plugin, hence the added interest)

Hi @navaronbracke  Thank you for reviewing the PR  .Appreciate you taking your time for the same

auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
```diff
@@ -1,15 +1,20 @@
-import 'dart:js_util';
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
 import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interface.dart';
 
+@JS('Object.keys')
+external JSArray<JSString> keys(final JSObject o);
```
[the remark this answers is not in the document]
pmathew92: I did initially add this in the js_interop class but the code never got executed when called from other files. Hence moved to respective files. Agreed this needs to be addressed

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -52,29 +52,29 @@ class RedirectLoginOptions {
 
 @JS()
 @anonymous
-abstract class Cache {
-  T get<T>(final String key);
-  void remove(final String key);
-  void set<T>(final String key, final T entry);
-  Future<List<String>> allKeys();
+extension type  Cache._(JSObject _) implements JSObject {
```
navaronbracke, line 55: There is an extra space here, after the `extension type ` declaration

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -52,29 +52,29 @@ class RedirectLoginOptions {
...
 }
 
 @JS()
 @anonymous
-class Auth0ClientInfo {
-  external Map<String, String>? get env;
-  external String get name;
+extension type Auth0ClientInfo._(JSObject _) implements JSObject {
+  external JSObject? get env;
+  external String  get name;
```
navaronbracke, line 66: There is an extra space here

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -148,15 +148,15 @@ class LogoutParams {
 
 @JS()
 @anonymous
-class LogoutOptions {
+extension type  LogoutOptions._(JSObject _) implements JSObject {
```
navaronbracke, line 151: There is an extra space here, after the `extension type ` declaration

auth0_flutter/lib/src/web/js_interop.dart
```diff
@@ -106,7 +106,7 @@ class GetTokenSilentlyAuthParams {
 
 @JS()
 @anonymous
-class GetTokenSilentlyOptions {
+extension type GetTokenSilentlyOptions._(JSObject _) implements JSObject {
   external GetTokenSilentlyAuthParams? get authorizationParams;
   external String? get cacheMode;
   external num? get timeoutInSeconds;
```
navaronbracke, line 112: This should also be a `JSNumber?` ? Maybe we should do a pass for the remaining `num` declarations.

auth0_flutter/lib/src/web/extensions/web_exception_extensions.dart
```diff
@@ -1,15 +1,20 @@
...
-    final Map<String, dynamic> details = {};
+    final error = jsException.toJSBox.getProperty<JSString>('error'.toJS);
+    final description =
+        jsException.toJSBox.getProperty<JSString>('error_description'.toJS);
+    final Map<String, JSAny?> details = {};
 
-    objectKeys(jsException).forEach((final key) {
-      if (key == 'error' || key == 'error_description') return;
-      details[key as String] = getProperty<dynamic>(jsException, key);
+    keys(jsException.toJSBox).toDart.forEach((final key) {
```
navaronbracke, line 15: Small suggestion to make this easier to read. We should also not mix up the JS types with the Dart types.
Each Dart string should be manually converted to a JSString. While a JSString _might_ be equal to a String (it is one of the allowed primitives), this might not always be the case and could be JS backend dependent in the future.

```suggestion
    keys(jsException.toJSBox).toDart.forEach((final JSString key) {
       if (key == 'error'.toJS || key == 'error_description'.toJS) return;
       details[key.toDart] = jsException.toJSBox.getProperty<JSAny?>(key);
    });
```

navaronbracke (changes requested):
Some minor follow-up comments, but this is looking good!

pmathew92:
Hi @navaronbracke  , Would you know what would be the best way to modify the test class 
https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/test/web/auth0_flutter_web_test.mocks.dart
Currently I am getting the error 
`Classes and mixins can only implement other classes and mixins.`
```class _FakeAuth0Client_0 extends _i1.SmartFake implements _i2.Auth0Client {
  _FakeAuth0Client_0(
    Object parent,
    Invocation parentInvocation,
  ) : super(
          parent,
          parentInvocation,
        );
}```
in the above class declaration as Auth0Client is now an extension type

**Comment by navaronbracke** [comment]
Extension type mocking for the new JS types requires some changes.

See the docs at https://dart.dev/interop/js-interop/mock

mockito 5.4.5 did add support for extension types, though, so you can als look into that.

**Comment by pmathew92** [comment]
Hi @navaronbracke  I have updated the PR and fixed the failing UTs
One query , we have this check in the web.dart file which will load the correct file when compiled to WASM
```export 'web/auth0_flutter_plugin_stub.dart'
    if (dart.library.js_interop) 'web/auth0_flutter_plugin_real.dart';```
 But this will fail for users who don't compile to WASM. Is there any way for this check to consider for non-WASM compiles ?

**Comment by navaronbracke** [comment]
`dart:html` does not work in the new style JS interop, which is why the import should be changed to `dart.library.js_interop`. The `dart:js_interop` import is available for both the JS backend & the WASM backend.

Since you updated the minimum Dart & Flutter versions in the pubspec, users that want to compile to JS or WASM, using WASM compatible API's are going to need to migrate to that minimum Flutter version.

So I don't see why this would fail if you are running on Flutter 3.24 and trying to compile to JS. Both JS & WASM should work with the new API's.

File: auth0_flutter_platform_interface/pubspec.yaml [hunk 11]
```diff
@@ -5,13 +5,14 @@ version: 1.8.0
 homepage: https://github.com/auth0/auth0-flutter
 
 environment:
-  sdk: ">=2.17.0 <4.0.0"
-  flutter: ">=3.0.0"
+  sdk: ">=3.5.0 <4.0.0"
+  flutter: ">=3.24.0"
```
Widcket, line 9: We will need to call out in the release notes this change.

Widcket: Is `3.24.0` the minimum Flutter version we can support?

auth0_flutter/test/web/extensions/web_exception_extension_test.dart
```diff
@@ -1,20 +1,19 @@
...
+  JSObject createJsException(final String error, final String description,
+      {final Map<String, JSAny>? additionalProps}) {
+    final jsObject = JSObject();
+    jsObject.setProperty('error'.toJS, error.toJS);
+    jsObject.setProperty('error_description'.toJS, description.toJS);
 
     if (additionalProps != null) {
       for (final element in additionalProps.entries) {
-        setProperty(jsObject, element.key, element.value);
+        jsObject.setProperty(element.key.toJS, element.value);
```
Widcket, line 16: Should we call `element.value.toJS` here, instead of the the call site in the tests?

auth0_flutter_platform_interface/pubspec.yaml
Widcket: Same for `auth0_flutter/pubspec.yaml`.

pmathew92: > Is 3.24.0 the minimum Flutter version we can support?

Yes, this is the minimum version required to run WASM compilation
https://docs.flutter.dev/platform-integration/web/wasm#switch-to-the-latest-version-of-flutter

pmathew92: > We will need to call out this change in the release notes.
Same for auth0_flutter/pubspec.yaml.

Agree. will add this in the release notes while publishing

auth0_flutter/test/web/extensions/web_exception_extension_test.dart
pmathew92: Updated

Widcket (approved):

PR #573, opened 2025-05-07
Added webAuth cancel for ios to prevent active transaction errors

pmathew92:
### Changes
This PR added support to cancel any ongoing transaction on iOS while using UL

### References
https://github.com/auth0/Auth0.swift/blob/master/FAQ.md#5-how-to-resolve-the-failed-to-start-this-transaction-as-there-is-an-active-transaction-at-the-moment-error


auth0_flutter/macos/Classes/WebAuth/WebAuthCancelMethodHandler.swift
```diff
@@ -0,0 +1 @@
+../../../darwin/Classes/WebAuth/WebAuthCancelMethodHandler.swift
```
NandanPrabhu, line 1: do we need these changes?

auth0_flutter/darwin/Classes/WebAuth/WebAuthHandler.swift
```diff
@@ -49,6 +49,13 @@ public class WebAuthHandler: NSObject, FlutterPlugin {
     }
 
     public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+
+        if call.method == "webAuth#cancel"  {
```
NandanPrabhu, line 53: suggest to have the call method name as part of string constants

auth0_flutter/FAQ.md
```diff
@@ -116,6 +117,32 @@ This library has no control whatsoever over the alert box. Its contents cannot b
...
+
+```dart
+ try {
+  await webAuth.login(useHTTPS: false);
+} catch (e) {
+if(e.toString() == 'Failed to start this transaction, as there is an active transaction at the moment '){
+webAuth.cancelWebAuth();
+//Retry login
+}
+}
```
Widcket, line 140: Check the formatting of the code sample. It hinders readability.

auth0_flutter/lib/src/mobile/web_authentication.dart
```diff
@@ -130,9 +130,23 @@ class WebAuthentication {
     await _credentialsManager?.clearCredentials();
   }
 
+
+  /// Terminates the ongoing web-based operation and reports back that it was cancelled.
+  /// You need to call this method within your custom Web Auth provider implementation whenever the operation is
+  /// cancelled by the user.
+  /// ## Note:
+  /// This is an iOS specific api
```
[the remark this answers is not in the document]
Widcket: Also, should it be a static method or not?

auth0_flutter/darwin/Classes/WebAuth/WebAuthHandler.swift
pmathew92: Adding this as part of the existing Method enum was failing the UTs as all the corresponding handlers expect arguments. Fixing that would require adding some explicit conditions in UTs and code which weren't looking good

auth0_flutter/macos/Classes/WebAuth/WebAuthCancelMethodHandler.swift
pmathew92: Do you mean the WebAuthCancelMethodHandler class ?

auth0_flutter/FAQ.md
pmathew92: done

auth0_flutter/lib/src/mobile/web_authentication.dart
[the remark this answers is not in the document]
Widcket: In this case, it depends on how it's going to be used. In the case of the Auth0.swift SDK, the `cancel()` method is meant to be called on its own (even possibly on the App Delegate) in circumstances where there is no expectation of an existing instance being around. Not sure what the usage would look like on Flutter apps, hence the question.

auth0_flutter/FAQ.md
Widcket: The `if` braces are at the same level as those from the `catch`, and the if block itself is not indented, where the contents of the `try` block are.

Screenshot 2025-05-07 at 15 00 34

auth0_flutter/FAQ.md
```diff
@@ -116,6 +117,33 @@ This library has no control whatsoever over the alert box. Its contents cannot b
...
+##### Clear the login transaction when handling the `transactionActiveAlready` error
+
+You can invoke `cancelWebAuth()` to manually clear the current login transaction upon encountering this error. Then, you can retry login. For example:
+
+```dart
+try {
+    await webAuth.login(useHTTPS: false);
+} catch (e) {
+    if (e.toString() ==
+    'UNKNOWN: Failed to start this transaction, as there is an active transaction at the moment.') {
```
Widcket, line 137: Let's avoid relying on error descriptions as much as possible, as they are not part of the public API.

In this case, a new error code could be added here: https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/darwin/Classes/WebAuth/WebAuthExtensions.swift#L12-L20

So the comparison suggested here relies only on the error code.

auth0_flutter/lib/src/mobile/web_authentication.dart
```diff
@@ -131,13 +131,11 @@ class WebAuthentication {
...
-  /// You need to call this method within your custom Web Auth provider implementation whenever the operation is
-  /// cancelled by the user.
-  /// ## Note:
-  /// This is an iOS specific API
+  /// Terminates the ongoing web-based operation and reports back that it was
+  /// cancelled.
+  /// ## Note: This is an iOS specific API
   ///
-  void cancelWebAuth() {
+ static void cancelWebAuth() {
```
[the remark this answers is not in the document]
pmathew92: Renamed to cancel

Widcket (approved):

NandanPrabhu (approved):

Issue #378, opened 2023-12-11
Allow accessing credentials without an attempt to refresh


MilesAdamson:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

For apps which are designed around the user being offline frequently, the following flow would be great:
- user is logged in
- user goes offline
- user returns to the app while offline, access token expired
- user can continue to browse the app with their identity in the state they last saw it
- upon coming back online, access token refresh happens

Currently, this is harder to achieve because `credentials` of `DefaultCredentialsManager` will always try to refresh the token, which fails without internet. Since this is the only method which access credentials, there is no way to just get the cached (expired) credentials. This would be convenient to have so that in the offline state, we can still restore the users id/roles/permissions as to what they should be able to view in that offline state.

### Describe the ideal solution

Add a method to `CredentialsManager` to get the cached credentials, without any attempt to refresh them. Or, add a boolean to the input of `CredentialsManager.credentials` to do the same behavior. 

### Alternatives and current workarounds

Alternatively it would be possible to cache the users id/roles/permissions ourselves to use in the app during the offline state. But this would be redundant with the credentials already in the manager, if we could access them.

It would also be possible to write our own implementation of credentials manager, to returns expired credentials from `credentials` if it fails to refresh while offline. But it defeats the purpose of the default manager to need to re-write the api calls to refresh the tokens and such

Widcket:
Hi @MilesAdamson, thanks for raising this.

To clarify, what you need is offline access to the _user profile_ only (the contents of the ID token), right?

[...]

PR #607, opened 2025-06-26
CredentialsManager user info/ID token contents accessible via flutter SDK

NandanPrabhu:

### 📋 Changes

This PR exposes api to access ID Token contents via Credentials Manager  in flutter SDK
GH issue https://github.com/auth0/auth0-flutter/issues/378 will be addressed via this PR

### DIFF
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -15,6 +15,7 @@
 - [📱 Credentials Manager](#-credentials-manager)
   - [Check for stored credentials](#check-for-stored-credentials)
   - [Retrieve stored credentials](#retrieve-stored-credentials)
+  - [Retrieve user profile](#retrieve-user-profile)
   - [Custom implementations](#custom-implementations)
   - [Local authentication](#local-authentication)
   - [Credentials Manager configuration](#credentials-manager-configuration)
@@ -395,6 +396,7 @@ await webAuth.logout();
 
 - [Check for stored credentials](#check-for-stored-credentials)
 - [Retrieve stored credentials](#retrieve-stored-credentials)
+- [Retrieve user profile](#retrieve-user-profile)
 - [Custom implementations](#custom-implementations)
 - [Local authentication](#local-authentication)
 - [Credentials Manager configuration](#credentials-manager-configuration)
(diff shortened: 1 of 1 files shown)


auth0_flutter_platform_interface/lib/src/user_info_credentials.dart
```diff
@@ -0,0 +1,111 @@
+class UserIdentity {
```
pmathew92, line 1: We already have a User_Profile class defined  . Please re use that

auth0_flutter/lib/src/mobile/credentials_manager.dart
```diff
@@ -9,7 +9,9 @@ abstract class CredentialsManager {
     final Map<String, String> parameters = const {},
   });
 
-  Future<bool> storeCredentials(final Credentials credentials);
+  Future<UserInfo> getIDTokenContents();
```
pmathew92, line 12: Shouldn't  this be of nullable type ,in the scneario if the credentials aren't available and someone invokes this api ?

auth0_flutter_platform_interface/lib/src/credentials-manager/credentials_manager_platform.dart
```diff
@@ -36,6 +35,11 @@ abstract class CredentialsManagerPlatform extends PlatformInterface {
     throw UnimplementedError('getCredentials() has not been implemented');
   }
 
+  /// Retrieves the credentials from the native storage.
+  Future<UserInfo> getIDTokenContents(final CredentialsManagerRequest request) {
```
Widcket, line 39: The `Credentials` class [uses](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter_platform_interface/lib/src/credentials.dart#L48) a `user` property for this. Since this method here is exposing exactly the same functionality (parsing and returning the info contained in the ID token), we should stick to the same naming convention here, otherwise it could be confusing (different names for the same thing may suggest these are actually different things, or that there is a substantive difference).

sanchitmehta94 (commented):
@NandanPrabhu Let’s resolve the open comments, then we can test and merge this to close the GH issue

pmathew92:
> @pmathew92 this is for both the platforms right?
@NandanPrabhu Yes

Issue #395, opened 2024-01-10
Consolidate `WebAuthenticationException`s caused by a user cancelling the flow into a single exception.


danallen88:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

Users cancelling a web authentication session workflow results in a `WebAuthenticationException` with an internal `code` field that describes the cause for the exception. Presumably this is because many things could cause this type of exception. 

I have a use case where knowing specifically that the user cancelled the session is helpful, but in order to identify whether this action was taken by a user, I must inspect the code for each platform that can throw it (iOS and Android for me) and determine whether the code is either `USER_CANCELLED` (iOS) or `a0.authentication_canceled` (Android).  These appear to be platform-specific codes that are emitted from a native layer.

### Describe the ideal solution

Ideally there is an exception type that wraps any scenario where a user cancelled the web authentication session, perhaps by subclassing `WebAuthenticationException` with a new exception `UserCancelledWebAuthenticationException`. Internal to the SDK, this exception can be created by inspecting the variety of codes that indicate the exception, and create a single exception that could be caught and dealt with.

### Alternatives and current workarounds

As it stands, this is the class I am using to capture these codes:

```dart
import 'package:auth0_flutter/auth0_flutter.dart'
    show WebAuthentication, WebAuthenticationException;

/// {@template web_authentication_exception_code}
/// Error codes that can be returned from a [WebAuthentication] session within a
/// [WebAuthenticationException].
///
/// These codes are emitted from the native layer which means that different
/// unique codes can exist for the same type of exception on different
[...]

PR #609, opened 2025-07-02
chore : added new properties to web_authentication_exception and credentials_manager_exception class

pmathew92:
### 📋 Changes

This PR adds new properties in the `web_authentication_exception` and `credentials_manager_exception` class to easily compare for specific error scenarios
 
### 📎 References
#395

### DIFF
--- a/auth0_flutter_platform_interface/lib/src/credentials-manager/credentials_manager_exception.dart
+++ b/auth0_flutter_platform_interface/lib/src/credentials-manager/credentials_manager_exception.dart
@@ -22,4 +22,19 @@ class CredentialsManagerException extends Auth0Exception {
 
     return CredentialsManagerException(e.code, e.messageString, errorDetails);
   }
+
+  bool get isTokenRenewFailed =>
+      code == 'RENEW_FAILED' ||
+      code ==
+          '''
+An error occurred while trying to use the Refresh Token to renew the Credentials.''';
+
+  bool get isNoCredentialsFound =>
+      code == 'NO_CREDENTIALS' || code == 'No Credentials were previously set.';
+
+  bool get isNoRefreshTokenFound =>
+      code == 'NO_REFRESH_TOKEN' ||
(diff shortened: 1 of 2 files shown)


auth0_flutter_platform_interface/lib/src/credentials-manager/credentials_manager_exception.dart
```diff
@@ -22,4 +22,19 @@ class CredentialsManagerException extends Auth0Exception {
 
     return CredentialsManagerException(e.code, e.messageString, errorDetails);
   }
+
+  bool get isTokenRenewFailed =>
+      code == 'RENEW_FAILED' ||
+      code ==
+          '''
+An error occurred while trying to use the Refresh Token to renew the Credentials.''';
```
Widcket, line 30: These need a proper error code on the Android side as well.

Widcket (approved):

NandanPrabhu:
@pmathew92 iOS checks in the CI are failing but merge button is enabled. Is it expected?

wiesnery:
Hi @pmathew92,
Thank you for your quick reply. I checked it and exception codes are different for iOS and Android. For Android, I digged through the code of the `SecureCredentialsManger` and around line 566, errors of all kind will just be wrapped to a kind of generic `CredentialsManagerException` with message = code "An error occurred while trying to use the Refresh Token to renew the Credentials." which makes it impossible to distinguish in the flutter code. This is although the underlying `AuthenticationException` is amazingly well with distinguishing error causes (maybe even a bit too specific for beginners like me).

Could you confirm my findings or prove me wrong (I prefer the latter xD) and/or give me a hint on how to proceed maybe not only on the exception handling level, but maybe what should be correctly done with my high level problem on when to clearCredentials() and when better not to?

Cheers!

pmathew92:
@grebdems  Yes, #607  will also be included in the coming versions

Issue #409, opened 2024-02-15
redirect_uri incorrect in `auth0.onLoad()`.


jwalton:
### Checklist

- [X] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

When I login with:

```dart
_auth0 = Auth0Web(domain, clientId);
_auth0.loginWithRedirect(
  redirectUrl: 'http://localhost:8080/callback",
);

_auth0.onLoad().then((final credentials) { ... }
```

This works as expected.  But, if I reload the page, the `onLoad()` attempts to do a "silent authentication" to recover my credentials and this fails.  The `onLoad()` results in a network request being sent that looks like:

```dart
https://xxxx.auth0.com/authorize?...&redirect_uri=https%3A%2F%2Flocalhost%3A8080&...
```

And this fails, with the logs reporting "Failed Silent Auth: Callback URL mismatch. https://localhost:8080 is not in the list of allowed callback URLs."

This is correct: that URL is not in my list of allowed callback URLs, and nor do I want it to be. But, `onLoad()` doesn't seem to accept a redirectUrl as a parameter.

You can work around this with:

```dart
_auth0.onLoad(parameters: {'redirect_uri': _redirectUrl}).then((final credentials) {...})
```

But, yuck.  It would be much nicer if the `Auth0Web` constructor accepted an optional redirectUri parameter like the [js version does](https://auth0.com/docs/libraries/auth0js#available-parameters).

### Reproduction

[...]

PR #624, opened 2025-08-04
Adding fix for OnLoad() for refresh redirecting

utkrishtsahu:
- [ X] All new/changed/fixed functionality is covered by tests (or N/A)
- [ X] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

Added redirectUrl and cacheLocation to the Auth0Web constructor: These optional parameters allow developers to configure the redirect URI and authentication state storage location once at initialization. This ensures consistency between loginWithRedirect and onLoad() calls.

Fixed onLoad() silent authentication: Previously, onLoad() could fail on page refresh due to a redirect_uri mismatch or a missing_transaction error.

### 📎 References
https://github.com/auth0/auth0-flutter/issues/409

### 🎯 Testing

The new functionality is covered by manual end-to-end testing.

### DIFF
--- a/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
+++ b/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
@@ -33,7 +33,6 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
   /// Thus clearing this object is not needed,
   /// as the actual state is managed across reloads,
   /// using the transaction manager.
-  // TODO: move the `appState` to the result of `onLoad/initialize`
   Object? _appState;
 
   @override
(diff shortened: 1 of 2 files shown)


auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -57,6 +56,7 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
 
         _appState = result.appState.dartify();
 
+        window.history.replaceState(null, '', window.location.pathname);
```
pmathew92, line 59: What is the need for this piece of code ?

utkrishtsahu: window.history.replaceState removes the temporary code and state parameters from the URL after a successful login. This prevents the SDK from trying to process the same login a second time on page refresh, which would cause an error. It's a cleanup step that ensures silent authentication works correctly on subsequent visits.

Issue #615, opened 2025-07-15
Inconsistent use of `invitationUrl` in web and mobile sign-in


rob-clother-saible:
### Checklist

- [x] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I am writing a Flutter app for web, iOS and Android. I have to use  `Auth0Web.loginWithRedirect()` for web sign-in / onboarding and `Auth0.webAuthentication().login()` for mobile authentication. Both have an argument called `invitationUrl`.

The mobile login breaks if the invitation URL is not the full URL (which I am building up in code as follows):

```
class Auth0OrganizationInvitation {
  final String organization;
  final String organizationName;
  final String invitation;
  Auth0OrganizationInvitation({
    required this.organization,
    required this.organizationName,
    required this.invitation
  });

  String invitationUrl(AppConfig appConfig) {
    return 'https://${appConfig.webDomain}/signinpage?invitation=$invitation&organization=$organization&organization_name=$organizationName';
  }
}
```

However, the web login will give back "Invitation not found or already used" if the full URL is used. Instead, the invitation ID should be passed to the `loginWithRedirect` function. This is inconsistent, and the argument name is misleading.

### Reproduction

1. Use the class above to encapsulate an organization invitation.
2. Implement mobile sign-in as follows:

```
      final credentials = await config.auth0.webAuthentication().login(
        useHTTPS: true,
        organizationId: invitation?.organization,
[...]

PR #645, opened 2025-08-28
Resolves the inconsistency where the web platform required an invitation ticket ID for the  parameter, while mobile platforms required the full URL.

utkrishtsahu:
- [X ] All new/changed/fixed functionality is covered by tests (or N/A)
- [ X] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

Resolves the inconsistency where the web platform required an invitation ticket ID for the `invitationUrl` parameter, while mobile platforms required the full URL.

This change introduces a helper function that checks if the provided `invitationUrl` is a full URL. If it is, it parses the `invitation` query parameter to extract the ticket ID. If it's not a full URL, it's assumed to be the ticket ID itself.

This makes the `invitationUrl` parameter behave consistently across all platforms, allowing developers to pass the full invitation URL without causing an "invitation not found" error on the web.

### 📎 References
https://github.com/auth0/auth0-flutter/issues/615

### 🎯 Testing

E2E testing for Web and added unit test case for the same.

### DIFF
--- a/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
+++ b/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
@@ -69,7 +69,7 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
   @override
   Future<void> loginWithRedirect(final LoginOptions? options) {
     final client = _ensureClient();
-    final invitationTicket = getInvitationTicket(options?.invitationUrl);
+    final invitationTicket = _getInvitationTicket(options?.invitationUrl);
 
     final authParams = JsInteropUtils.stripNulls(JsInteropUtils.addCustomParams(
         interop.AuthorizationParams(
@@ -94,7 +94,7 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
   @override
   Future<Credentials> loginWithPopup(final PopupLoginOptions? options) async {
     final client = _ensureClient();
-    final invitationTicket = getInvitationTicket(options?.invitationUrl);
+    final invitationTicket = _getInvitationTicket(options?.invitationUrl);
 
     final authParams = JsInteropUtils.stripNulls(JsInteropUtils.addCustomParams(
         interop.AuthorizationParams(
@@ -154,7 +154,7 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
   @override
   Future<bool> hasValidCredentials() => clientProxy!.isAuthenticated();
 
-  String? getInvitationTicket(final String? invitationUrl) {
+  String? _getInvitationTicket(final String? invitationUrl) {
     if (invitationUrl == null || invitationUrl.isEmpty) {
       return null;
     }
--- a/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
+++ b/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
@@ -39,18 +39,21 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
   Future<Object?> get appState => Future<Object?>.value(_appState);
 
   String? _getInvitationTicket(final String? invitationUrl) {
-    if (invitationUrl == null) {
+    if (invitationUrl == null || invitationUrl.isEmpty) {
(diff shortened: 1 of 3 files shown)


auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -38,6 +38,21 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
   @override
   Future<Object?> get appState => Future<Object?>.value(_appState);
 
+  String? _getInvitationTicket(final String? invitationUrl) {
```
pmathew92, line 41: Why is the method name starting with `_`. It is a public facing one isn't it ?

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -38,6 +38,21 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
...
+    if (invitationUrl == null) {
+      return null;
+    }
+
+    final uri = Uri.tryParse(invitationUrl);
+    // If it's a full URL, parse the 'invitation' query parameter.
+    if (uri != null && uri.hasAuthority) {
+      return uri.queryParameters['invitation'];
```
pmathew92, line 49: What if the uri doesn't have an `invitation` parameter ?

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -413,4 +413,91 @@ void main() {
             e.code == 'test' &&
             e.message == 'test exception')));
   });
+
```
pmathew92, line 416: Add scenarios to handle when we pass and invalid url , an empty string , and a valid url without the `invitation` query parameter

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
utkrishtsahu: _getInvitationTicket is an internal helper function for this class only and not a public-facing API

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
utkrishtsahu: In Dart, accessing a map key that doesn't exist (like uri.queryParameters['invitation']) returns null instead of throwing an error. Since the method's return type is String?, this was the intended behavior.
Also added the test case for the same where the URL is valid but the invitation parameter is missing.

auth0_flutter/test/web/auth0_flutter_web_test.dart
utkrishtsahu: Added test scenarios for the same.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
pmathew92: Declare this as private . I don't think we are using the `_` prefix on any of the method so far. Lets keep it that way

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -386,31 +385,145 @@ void main() {
...
+              () async {
+            await auth0.loginWithRedirect(invitationUrl: invitationId);
+
+            final captured = verify(mockClientProxy.loginWithRedirect(captureAny))
+                .captured
+                .single as interop.RedirectLoginOptions;
+            expect(captured.authorizationParams!.invitation, invitationId);
+          });
+
+      test('treats an invalid URL as a ticket ID', () async {
```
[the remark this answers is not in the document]
utkrishtsahu: test name was imprecise. The code's logic is designed to fall back and use the original string if URL parsing fails, which is what the test was asserting.
 renamed the test to **'uses the original string as ticket ID when URL parsing fails'** and added a comment explaining this exact behavior.

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
utkrishtsahu: Done

PR #393, opened 2023-12-22
Feature/add openUrl support

JCQuintas:
- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [x] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

- Adds support for the `openUrl` parameter in both `loginWithRedirect` and `logout` on the browser version.
- `openUrl` requires *auth0-spa-js* version `2.0.1` or later.
- It makes the lib more closely aligned with the *auth0-spa-js* lib which is used underneath.

This is important to me due to the [@microsoft/teams-js](https://github.com/OfficeDev/microsoft-teams-library-js) library necessary to do authentication inside microsoft teams, [reference docs here](https://learn.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/authentication/auth-tab-aad?tabs=teamsjs-v2)

In short, it handles authentication in its own window. 

```ts
authentication.authenticate({
    url: window.location.origin + "/tab-auth/simple-start-v2",
    width: 600,
    height: 535})
```

My use case would be something similar to 

```dart
auth0.loginWithRedirect(openUrl: (auth0Url) async {
  authentication.authenticate(
    url: "$auth0Url/tab-auth/simple-start-v2",
    width: 600,
    height: 535
  )
})
```

### 🎯 Testing

- In order to test, replace the auth0-spa-js version from `2.0` to `2.0.1` at `auth0_flutter/example/web/index.html`
- Add the snippet below to `loginWithRedirect` and `logout` functions
```dart
openUrl: (final url) async {
  debugPrint(url);
  window.location.replace(url);
}
```
- Ensure you will see the urls printed on the console.

## Possible concerns

While manual testing, you will also two errors:

- `Error: Promise was rejected with a value of 'undefined'`
- `TypeError: T.as is not a function`
  
From my tests, this seems be comming out of the main `auth0-spa-js` library due to unhandled rejections, and doesn't affect the usage of `auth0_flutter` itself. Though it would require a deeper research on why it is happening.

PR #685, opened 2025-11-24
Feature/add openurl support

trevordunn:
This is a continuation of #393, which seems to have gone stale.

* [x]  All new/changed/fixed functionality is covered by tests (or N/A)
* [x]  I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes
* Adds support for the `openUrl` parameter in both `loginWithRedirect` and `logout` on the browser version.
* `openUrl` requires _auth0-spa-js_ version `2.0.1` or later.
* It makes the lib more closely aligned with the _auth0-spa-js_ lib which is used underneath.

This is important to me due to the [@microsoft/teams-js](https://github.com/OfficeDev/microsoft-teams-library-js) library necessary to do authentication inside microsoft teams, [reference docs here](https://learn.microsoft.com/en-us/microsoftteams/platform/tabs/how-to/authentication/auth-tab-aad?tabs=teamsjs-v2)

In short, it handles authentication in its own window.

```ts
authentication.authenticate({
    url: window.location.origin + "/tab-auth/simple-start-v2",
    width: 600,
    height: 535})
```

My use case would be something similar to

```dart
auth0.loginWithRedirect(openUrl: (auth0Url) async {
  authentication.authenticate(
    url: "$auth0Url/tab-auth/simple-start-v2",
    width: 600,
    height: 535
  )
})
```

### 🎯 Testing
* In order to test, replace the auth0-spa-js version from `2.0` to `2.0.1` at `auth0_flutter/example/web/index.html`
* Add the snippet below to `loginWithRedirect` and `logout` functions

```dart
openUrl: (final url) async {
  debugPrint(url);
  window.location.replace(url);
}
```

* Ensure you will see the urls printed on the console.

## Possible concerns
While manual testing, you will also see two errors in the web console:

* `Error: Promise was rejected with a value of 'undefined'`
* `TypeError: T.as is not a function`

From my tests, this seems be comming out of the main `auth0-spa-js` library due to unhandled rejections, and doesn't affect the usage of `auth0_flutter` itself. Though it would require a deeper research on why it is happening.

### DIFF
--- a/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
+++ b/auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
@@ -85,11 +85,11 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
 
     final openUrl = options?.openUrl;
 
-    final loginOptions = interop.RedirectLoginOptions(
+    final loginOptions = JsInteropUtils.stripNulls(interop.RedirectLoginOptions(
       appState: options?.appState.jsify(),
       authorizationParams: authParams,
       openUrl: openUrl,
-    );
+    ));
 
     return client.loginWithRedirect(loginOptions);
   }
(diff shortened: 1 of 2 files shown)


trevordunn:
Stalled on finishing this until #686 is resolved.

pmathew92:
Hi @trevordunn, Just checking upon this . Are you facing any issues now that #686  is fixed ?

trevordunn:
@pmathew92 I'm still seeing #686. Left a [comment](https://github.com/auth0/auth0-flutter/issues/686#issuecomment-3729913331) there.

Issue #686, opened 2025-11-24
Build runner throws "Bad state: Asset URI is missing for abstract class JSObject" error


trevordunn:
### Checklist

- [x] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I've run these commands on `main`:
```
cd auth0_flutter
dart run build_runner clean
dart run build_runner build --delete-conflicting-outputs
```

And getting this output:
```
13s mockito:mockBuilder on 11 inputs: 3 output, 1 no-op; spent 9s analyzing, 4s sdk, 1s resolving; test/web/auth0_flutter_web_test.dart                                
                                                                                                                                                                       
Building, full build because builders changed.                                                                                                                         
                                                                                                                                                                       
log output for mockito:mockBuilder on test/web/auth0_flutter_web_test.dart                                                                                             
E Bad state: Asset URI is missing for abstract class JSObject                                                                                                          
  #0      _MockClassInfo._typeImport (package:mockito/src/builder.dart:2524:10)                                                                                        
  #1      _MockClassInfo._addFakeClass.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:mockito/src/builder.dart:1913:25)                          
  #2      _$TypeReferenceBuilder.update (package:code_builder/src/specs/type_reference.g.dart:164:33)                                                                  
  #3      new _$TypeReference (package:code_builder/src/specs/type_reference.g.dart:22:36)                                                                             
[...]

PR #708, opened 2025-12-19
fix: Fixes the build error when running the `dart run build_runner build --delete-conflicting-outputs` command

pmathew92:
### 📋 Changes
This PR fixes the build error when running the `dart run build_runner build --delete-conflicting-outputs` command. 
The command deletes the generated mock file for the `auth0_flutter_web_test.dart` file and fails to re-generae it which resulted in the error

### 📎 References

#686 

### 🎯 Testing
This was tested manually to ensure this compiles successfully


utkrishtsahu (approved):
LGTM

PR #749, opened 2026-02-25
feat: Implement Native to Web

sanchitmehtagit:
- [x] All new/changed/fixed functionality is covered by tests (or N/A)
- [x] I have added documentation for all new/changed functionality (or N/A)

  ### Native to Web SSO 

  Mobile apps and web apps typically maintain separate authentication sessions. When a user is already authenticated in your native iOS or Android app and taps a link that opens a web page (in a browser or WebView), they are presented with a login screen again — even though they just logged in. This creates a jarring, broken experience.

  Native to Web SSO solves this by allowing the native app to transfer its authenticated session to the web context. Instead of the user re-authenticating, the app exchanges its stored refresh token for a short-lived, single-use session transfer token. That token is appended to the URL when opening the web app, which then uses it to establish an Auth0 session automatically — no second login prompt.

### 📋 Changes

1.   Adds ssoCredentials() to CredentialsManager, enabling Native to Web SSO — the ability to hand off an authenticated native session to a web application without requiring the user to log in again.
2.   The method exchanges the stored refresh token for a short-lived, single-use SessionTransferCredentials containing a sessionTransferToken. The token is passed to the web app as a session_transfer_token query parameter or cookie, where it can be forwarded to Auth0's /authorize endpoint to establish a web session.

  #### Dart / Platform Interface

  - Added SessionTransferCredentials model with sessionTransferToken, tokenType, expiresIn, idToken?, and refreshToken? fields
  - Added GetSSOCredentialsOptions with parameters and headers support
  - Added getSSOCredentials to CredentialsManagerPlatform and MethodChannelCredentialsManager
  - Added ssoCredentials() to CredentialsManager with full API documentation including Early Access notice, token characteristics, prerequisites, and See also link

  #### Android

  - New GetSSOCredentialsRequestHandler — delegates to SecureCredentialsManager.getSsoCredentials(parameters, callback)
  - Note: headers is not forwarded as the underlying auth0.android SDK does not expose a headers parameter on getSsoCredentials

  #### iOS / macOS (darwin)

  - New CredentialsManagerSSOCredentialsMethodHandler — delegates to CredentialsManager.ssoCredentials(parameters:headers:)
  - Registered in CredentialsManagerHandler and symlinked across ios/ and macos/

### 📎 References

- https://auth0.com/docs/authenticate/single-sign-on/native-to-web/configure-implement-native-to-web — setup guide for enabling the feature on your tenant and configuring native applications

#### Manual Testing in Flutter App
[...]

PR #759, opened 2026-03-02
feat: add ssoExchange to AuthenticationApi for Native to Web SSO

sanchitmehtagit:
## Context
 
 PR #749 introduced Native to Web SSO via CredentialsManager.ssoCredentials(). That works well for apps that use CredentialsManager to store and manage credentials. However, apps that manage tokens and their rotation themselves and keep the refresh token in their own storage - had no path to Native to Web SSO..

## Changes 
  **What this PR does**

  Adds AuthenticationApi.ssoExchange(refreshToken:) as a direct API-level method that exchanges a refresh token for SSOCredentials without requiring CredentialsManager.

  **Dart**
  - New AuthSSOExchangeOptions (channel options class with refreshToken, parameters, headers)
  - New ssoExchange() abstract method on Auth0FlutterAuthPlatform
  - MethodChannelAuth0FlutterAuth.ssoExchange() dispatches over auth#ssoExchange
  - AuthenticationApi.ssoExchange(refreshToken:, parameters:, headers:) as the public API

  **Android**
  - SSOExchangeApiRequestHandler calling AuthenticationAPIClient.ssoExchange(refreshToken) with addParameters() and per-entry addHeader()

  **iOS / macOS**
  - SSOExchangeMethodHandler calling Authentication.ssoExchange(withRefreshToken:).parameters().headers()
  - Registered as auth#ssoExchange case in AuthAPIHandler

**Refactorng**
  - Renamed CredentialsManagerSSOCredentialsMethodHandler → SSOCredentialsMethodHandler (removes redundant folder-prefix from struct name)

## Testing

  Tests
  - SSOExchangeApiRequestHandlerTest.kt — 8 Android unit tests (missing token, correct call, parameters, headers, success/failure mapping)
  - 6 new tests in method_channel_auth0_flutter_auth_test.dart — method channel key, argument mapping, defaults, response parsing, null and exception handling
  - 2 new tests in authentication_api_test.dart — pass-through and default values

### DIFF
--- a/auth0_flutter/example/ios/Tests/CredentialsManager/CredentialsManagerSSOCredentialsMethodHandlerTests.swift
+++ b/auth0_flutter/example/ios/Tests/CredentialsManager/CredentialsManagerSSOCredentialsMethodHandlerTests.swift
@@ -5,7 +5,7 @@ import Auth0
 
 fileprivate typealias Argument = SSOCredentialsMethodHandler.Argument
 
-class SSOCredentialsMethodHandlerTests: XCTestCase {
+final class SSOCredentialsMethodHandlerTests: XCTestCase {
     var spyAuthentication: SpyAuthentication!
     var spyStorage: SpyCredentialsStorage!
     var sut: SSOCredentialsMethodHandler!


auth0_flutter/darwin/Classes/AuthAPI/SSOExchangeMethodHandler.swift
```diff
@@ -0,0 +1,50 @@
+import Auth0
+
+#if os(iOS)
+import Flutter
```
sanchitmehtagit: add iOS test cases

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/api/SSOExchangeApiRequestHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+import com.auth0.android.callback.Callback
+import com.auth0.android.result.SSOCredentials
+import com.auth0.auth0_flutter.request_handlers.MethodCallRequest
+import com.auth0.auth0_flutter.toMap
+import com.auth0.auth0_flutter.utils.assertHasProperties
+import io.flutter.plugin.common.MethodChannel
+
+private const val AUTH_SSO_EXCHANGE_METHOD = "auth#ssoExchange"
+
+class SSOExchangeApiRequestHandler : ApiRequestHandler {
```
NandanPrabhu: final class

auth0_flutter/example/ios/Tests/CredentialsManager/CredentialsManagerSSOCredentialsMethodHandlerTests.swift
```diff
@@ -3,24 +3,24 @@ import Auth0
 
 @testable import auth0_flutter
 
-fileprivate typealias Argument = CredentialsManagerSSOCredentialsMethodHandler.Argument
+fileprivate typealias Argument = SSOCredentialsMethodHandler.Argument
 
-class CredentialsManagerSSOCredentialsMethodHandlerTests: XCTestCase {
+class SSOCredentialsMethodHandlerTests: XCTestCase {
```
NandanPrabhu: final class

auth0_flutter/example/ios/Tests/AuthAPI/AuthAPISSOExchangeMethodHandlerTests.swift
```diff
@@ -0,0 +1,108 @@
...
+        spy.ssoCredentialsResult = .success(ssoCredentials)
+        sut.handle(with: arguments()) { result in
+            let values = result as? [String: Any]
+            XCTAssertEqual(values?[SSOCredentialsProperty.refreshToken.rawValue] as? String, "new-refresh-token")
+            expectation.fulfill()
+        }
+        wait(for: [expectation])
+    }
+
+    func testDoesNotIncludeRefreshTokenWhenAbsent() {
```
pmathew92: Shouldn't the expectation in this case be to throw an error ?

auth0_flutter/EXAMPLES.md
```diff
@@ -626,7 +643,8 @@ await webViewController.loadRequest(
 >
 > Cookie injection is platform-specific and may require additional WebView configuration.
 
-
+> [!NOTE]
+> This feature is designed for **iOS and Android** only. Since the purpose is to transition a native app session into a web session, calling `auth0.credentialsManager.ssoCredentials()` or `auth0.api.ssoExchange()` from any other platform is not supported.
```
pmathew92: SPA js has a PR for N2W , so we might need to support it sometime later. Just update the note mentioning currently supported only in mobile

sanchitmehtagit: N2W APIs to fetch SSO is mobile only. on JS side they have raised a PR to support injecting SSO token as query paramter - https://github.com/auth0/auth0-spa-js/pull/1547

NandanPrabhu:
as flutter sdk supports web, would it make sense to add a feature to inject the sso token to web

sanchitmehtagit:
> as flutter sdk supports web, would it make sense to add a feature to inject the sso token to web

This is a nice Thought, but we need to evaluate whether there’s a valid use case at this point.

Currently, the expectation is that they will use the Flutter SDK as the mobile client (Android/iOS) to exchange for the SSO token (N2W is intended for the mobile client). After that, they would invoke a Flutter WebView to load their web app and inject the SSO token there.

Since on SPA JS, there’s already an open PR addressing a similar purpose, I’ll connect with Bharath to understand whether this use case is actually required going forward and whether it would meaningfully improve the developer experience

PR #804, opened 2026-04-02
feat: Swift Package Manager dependency support for Flutter iOS/macOS SDK

NandanPrabhu:

### 📋 Changes

Adds **Swift Package Manager (SPM)** support to the iOS/macOS (`darwin`) native plugin, addressing #793. **CocoaPods support is retained** as a fallback, so this is non-breaking for existing consumers.

**SPM packaging**
- Adds `auth0_flutter/darwin/auth0_flutter/Package.swift` declaring the `auth0_flutter` SPM target (iOS 14 / macOS 11), depending on Auth0.swift ≥ 2.21.2, JWTDecode.swift 3.3.0, and SimpleKeychain 1.3.0.
- Sets `sharedDarwinSource: true` in `pubspec.yaml` so Flutter resolves iOS and macOS from a single source tree under `darwin/auth0_flutter/Sources/auth0_flutter/`.
- Adds `PASSKEYS_PLATFORM` Swift compilation condition via `swiftSettings` in `Package.swift` (previously only provided by the CocoaPods podspec).

**CocoaPods fallback (retained)**
- `auth0_flutter/darwin/auth0_flutter.podspec` updated to point `source_files` at `Sources/auth0_flutter/` — the canonical location shared with SPM.
- `ios/` and `macos/` podspecs kept for CocoaPods consumers; their `Classes/` directories contain symlinks into `darwin/Classes/`, which in turn symlink to `Sources/auth0_flutter/`.
- A `scripts/generate-symlinks.sh` script regenerates the full symlink chain (`darwin/Classes/ → Sources/`, `ios/Classes/ → darwin/Classes/`, `macos/Classes/ → darwin/Classes/`) and is enforced by the `Symlinks` CI job on every PR.

**Plugin entry point**
- Removes the Objective-C shim (`Auth0FlutterPlugin.h/.m`) and renames `SwiftAuth0FlutterPlugin` → `Auth0FlutterPlugin` to match the `pluginClass` declared in `pubspec.yaml`. No public Dart API change.

**CI**
- `setup-darwin`: sets up Xcode before `flutter pub get`, enables SPM (`flutter config --enable-swift-package-manager`), caches and installs CocoaPods, then patches the Flutter-generated `FlutterGeneratedPluginSwiftPackage/Package.swift` deployment targets to 14.0 (iOS) / 11.0 (macOS).
- `unit-tests-darwin`: runs `xcodebuild test` against `Runner.xcworkspace` (CocoaPods workspace, which includes the SPM-resolved plugin).
- iOS coverage converted from `.xcresult` to Cobertura XML via `xcresultparser` before the Codecov upload.
- Adds `.github/workflows/check-symlinks.yml` CI job that runs `scripts/generate-symlinks.sh` and fails if any symlink is out of sync.

**Tests**
- Plugin registration tests renamed to `Auth0FlutterPluginTests`.
- Credentials Manager tests updated to use `credentialsManager.store(credentials:)` instead of hand-built `NSKeyedArchiver` fixtures.

### 📎 References

- Fixes #793
- Flutter SPM for plugin authors: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors

### 🎯 Testing

- iOS and macOS native unit tests pass in CI via `xcodebuild` against `Runner.xcworkspace` (Xcode 26.2); Android and Windows unit suites unchanged.
- The `Symlinks` CI job validates that `darwin/Classes/`, `ios/Classes/`, and `macos/Classes/` symlinks are consistent with `Sources/auth0_flutter/` on every PR.
- Manual verification: from `auth0_flutter/example`, run `flutter config --enable-swift-package-manager` then `flutter run` on an iOS simulator or device; exercise login/logout, credentials manager, and DPoP flows.
- CocoaPods fallback can be verified by running `flutter config --no-enable-swift-package-manager` and building the example app.

> **Reviewer note:** this PR is large by file count because `darwin/Classes/` files changed from full Swift source to single-line symlinks — git counts the old content as deletions and the canonical files under `Sources/auth0_flutter/` as additions. The substantive surface is `Package.swift`, the podspecs, the CI actions, and the plugin/test renames.


sanchitmehtagit:
@coderabbitai review

auth0_flutter/macos/Classes/AuthAPI/AuthAPIMultifactorChallengeMethodHandler.swift
```diff
@@ -1 +0,0 @@
-../../../darwin/Classes/AuthAPI/AuthAPIMultifactorChallengeMethodHandler.swift
```
sanchitmehtagit, line 1: do we need to delete this file?

NandanPrabhu:
@coderabbitai review

auth0_flutter/macos/Classes/AuthAPI/AuthAPIMultifactorChallengeMethodHandler.swift
NandanPrabhu: no. Its reinstated

.github/workflows/main.yml
```diff
@@ -13,7 +13,6 @@ concurrency:
     cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 env:
-    ruby: '3.3.1'
```
NandanPrabhu, line 16: ruby was primarily added cocoapods support. Github actions are updated to support swift package manager

NandanPrabhu:
@coderabbitai full review

scripts/generate-symlinks.sh
```diff
@@ -1,50 +1,80 @@
...
+
+    darwin_link="$darwin_classes_dir/$rel"
+    darwin_link_dir=$(dirname "$darwin_link")
+    mkdir -p "$darwin_link_dir"
+
+    # Depth of the symlink inside darwin/Classes (e.g. AuthAPI/X.swift -> depth 2)
+    link_depth=$(echo "$darwin_link" | awk -F '/' '{print NF-1}')
+    # We need to go up (link_depth) levels then down into auth0_flutter/Sources/auth0_flutter/
+    up=$(printf '../%.0s' $(seq 1 "$link_depth"))
+    ln -sf "${up}auth0_flutter/Sources/auth0_flutter/$rel" "$darwin_link"
```
utkrishtsahu, line 54: he up calculation is off here. link_depth is counting from the full path (auth0_flutter/darwin/Classes/...), so we end up with too many ../ and the link points at auth0_flutter/Sources/auth0_flutter/... instead of auth0_flutter/darwin/auth0_flutter/Sources/.... I pulled the branch and every link under darwin/Classes/ is dangling. Can you base the depth on the path relative to darwin/Classes/ instead?

scripts/generate-symlinks.sh
```diff
@@ -1,50 +1,80 @@
...
+
+        platform_link="$platform_classes_dir/$rel"
+        platform_link_dir=$(dirname "$platform_link")
+        mkdir -p "$platform_link_dir"
+
+        # Symlink ios/Classes/<rel> -> ../../../darwin/Classes/<rel>
+        # Depth of the link inside ios/Classes/
+        link_depth=$(echo "$platform_link" | awk -F '/' '{print NF-1}')
+        up=$(printf '../%.0s' $(seq 1 "$link_depth"))
+        ln -sf "${up}darwin/Classes/$rel" "$platform_link"
```
utkrishtsahu: Same problem as above down here — the ios/macos links resolve to <root>/darwin/Classes/... and miss the auth0_flutter/ prefix, so they're all broken too. I counted 183 dangling symlinks total across darwin/ios/macos. Worth adding a quick test that actually resolves the links after generating them.

auth0_flutter/darwin/auth0_flutter.podspec
```diff
@@ -11,7 +11,8 @@ Pod::Spec.new do |s|
   s.license      = { :file => '../LICENSE' }
   s.author       = { 'Auth0' => 'support@auth0.com' }
   s.source       = { :path => '.' }
-  s.source_files = 'Classes/**/*'
+  s.source_files = 'auth0_flutter/Sources/auth0_flutter/**/*.{h,m,swift}'
+  s.public_header_files = 'auth0_flutter/Sources/auth0_flutter/include/**/*.h'
```
utkrishtsahu: There's no include/ folder (or any .h/.m files) under Sources/auth0_flutter/ anymore since we dropped the ObjC shim, so this public_header_files line and the .{h,m} part of source_files above will just make CocoaPods warn about missing files. Can we trim these down to Swift-only?

.github/workflows/main.yml
```diff
@@ -151,15 +149,21 @@ jobs:
...
 
-            - name: Convert coverage report
+            # Codecov cannot parse a raw .xcresult bundle, so convert it to
+            # Cobertura XML first. xcresultparser is Ruby-free, keeping the SPM
+            # migration's goal of dropping the Ruby/slather toolchain.
+            - name: Convert xcresult to Cobertura coverage
               working-directory: auth0_flutter/example/ios
-              run: bundle exec slather coverage -x --scheme Runner Runner.xcodeproj
+              run: |
+                brew install a7ex/homebrew-formulae/xcresultparser
```
utkrishtsahu: This installs xcresultparser unpinned from a third-party tap on every run — bit risky for reproducibility and supply chain, and it'll break coverage silently if the CLI flags change upstream. Could we pin a version?

auth0_flutter/example/pubspec.yaml
```diff
@@ -57,8 +57,8 @@ flutter:
   # the material Icons class.
   uses-material-design: true
 
-  # assets:
-  #   - .env
+  assets:
+    - .env
```
utkrishtsahu: Uncommenting the .env asset means the build now hard-fails if .env doesn't exist. CI creates it so it's fine there, but a contributor running flutter run without one will hit an error instead of a warning. Intended?

auth0_flutter/darwin/auth0_flutter/Package.swift
```diff
@@ -0,0 +1,30 @@
...
+    name: "auth0_flutter",
+    platforms: [
+        .iOS("14.0"),
+        .macOS("11.0"),
+    ],
+    products: [
+        .library(name: "auth0-flutter", targets: ["auth0_flutter"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/auth0/Auth0.swift", .upToNextMajor(from: "2.21.2")),
```
utkrishtsahu: SPM uses .upToNextMajor(from: "2.21.2") for Auth0.swift but the podspecs pin 2.21.2 exactly, and Package.resolved already moved to 2.22.0. So the SPM and CocoaPods paths can end up on different versions. Fine if that's on purpose, just flagging.

utkrishtsahu (changes requested):

auth0_flutter/example/pubspec.yaml
NandanPrabhu: good catch. This has to be reverted

auth0_flutter/macos/Classes/MethodHandler.swift
```diff
@@ -1 +1 @@
-../../darwin/Classes/MethodHandler.swift
\ No newline at end of file
+../../../darwin/Classes/MethodHandler.swift
```
sanchitmehtagit: I would revert all the path-related changes across all files. There are a lot of them, and I'd remove those changes from this PR.

sanchitmehtagit (changes requested):

auth0_flutter/ios/Classes/MyAccountAPI/MyAccountEnrollPasskeyMethodHandler.swift
```diff
@@ -1 +1 @@
-../../../darwin/Classes/MyAccountAPI/MyAccountEnrollPasskeyMethodHandler.swift
\ No newline at end of file
+../../../../darwin/Classes/MyAccountAPI/MyAccountEnrollPasskeyMethodHandler.swift
```
sanchitmehtagit: Could we also verify if these symlink changes are intended. it seems to be broken currently as per my testing

scripts/generate-symlinks.sh
NandanPrabhu: Good catch, fixed in 4dc9e767. `link_depth` now counts subdirectories relative to `Classes/` (via `rel`, which is already stripped of the `sources_dir` prefix) instead of the full repo path, so the `../` count matches the actual nesting. Re-ran the script and verified 0 dangling symlinks across darwin/ios/macos (also added a post-generation resolution check to the script itself, see lines 82-94, so this class of bug fails CI instead of shipping silently).

scripts/generate-symlinks.sh
NandanPrabhu: Same root cause as the darwin/Classes issue above, fixed in the same commit (4dc9e767) — the ios/macos loop now derives `link_depth` from `rel` (relative to `Classes/`) instead of the full path, so `up` correctly resolves to `<platform>/../darwin/Classes/<rel>`. Also added the post-generation dangling-symlink check you suggested (script lines 82-94) — it walks all three Classes/ trees and fails the script (and CI, via the Symlinks job) if any link doesn't resolve.

auth0_flutter/darwin/auth0_flutter.podspec
NandanPrabhu: Trimmed in 4dc9e767 — `source_files` is now `auth0_flutter/Sources/auth0_flutter/**/*.swift` and `public_header_files` was removed entirely, since there's no ObjC shim or include/ dir left. Confirmed no .h/.m files remain under Sources/auth0_flutter/.

.github/workflows/main.yml
NandanPrabhu: Pinned in 4dc9e767 — now `brew install a7ex/homebrew-formulae/xcresultparser@2.0.1` instead of installing latest.

auth0_flutter/darwin/auth0_flutter/Package.swift
NandanPrabhu: This is intentional — Package.swift uses `.upToNextMajor(from: "2.21.2")` so SPM can pick up compatible patch/minor releases (like 2.22.0 in Package.resolved), and the podspecs were relaxed to `~> 2.21` in the same commit (4dc9e767) so CocoaPods tracks the same compatible range instead of being pinned to an exact version that would drift out of sync with SPM.

auth0_flutter/macos/Classes/MethodHandler.swift
NandanPrabhu: The path/symlink restructuring is required for this PR: SPM needs a single canonical source tree (`darwin/auth0_flutter/Sources/auth0_flutter/`), and CocoaPods (ios/, macos/, darwin/Classes/) needs to keep working off the same files without duplication, so the symlinks are what let both dependency managers point at one source of truth. What you were hitting was a real bug though — `scripts/generate-symlinks.sh` miscalculated the `../` depth and left ~185 symlinks dangling across darwin/ios/macos. That's fixed in 4dc9e767: the depth is now computed relative to `Classes/` instead of the full repo path, and I re-ran the script and verified every symlink resolves (0 dangling). I also added a post-generation check to the script itself so this can't silently regress again. Given the alternative (duplicating ~60 Swift files three times, or dropping CocoaPods support entirely) I'd like to keep the symlink approach — let me know if you still see broken links after pulling 4dc9e767.

auth0_flutter/ios/Classes/MyAccountAPI/MyAccountEnrollPasskeyMethodHandler.swift
NandanPrabhu: Confirmed intended and now verified working — this was the same depth-calculation bug reported by @utkrishtsahu on `scripts/generate-symlinks.sh`, fixed in 4dc9e767. I re-ran `scripts/generate-symlinks.sh` against current `main`-rebased HEAD and confirmed 0 dangling symlinks across `darwin/Classes/`, `ios/Classes/`, and `macos/Classes/` (including this file). Could you re-pull and re-test? If it's still broken for you I'd like to see the specific symlink/error so I can repro.

auth0_flutter/darwin/auth0_flutter/Package.swift
NandanPrabhu: Update: on reflection, an exact pin is safer here — a floating range can silently pick up a new Auth0.swift release between CI runs and cause SPM/CocoaPods to drift again. Pinned both to `2.22.0` exactly in e81a3572 (matches the version already locked in the checked-in `Package.resolved` files under example/ios and example/macos).

NandanPrabhu:
Closing in favor of a split, cascading pair of PRs — the diff here mixed a pure file-move/rename (moving `darwin/Classes/*` into a canonical `Sources/` tree, renaming `SwiftAuth0FlutterPlugin` → `Auth0FlutterPlugin`) with the actual SPM-specific work, which is most of why this ballooned past CodeRabbit's 150-file review limit.

- #894 — extracts the pure move/rename (no behavior change, byte-identical file moves)
- #893 — the SPM support itself, stacked on #894

All review feedback from this PR (podspec version bump, case-sensitive platform comparisons, symlink depth bug, unused `.h`/`.m` glob entries, xcresultparser pinning, `.env` asset revert, Auth0.swift version pinning) carries forward into the new PRs — most of it was already fixed on this branch and is preserved; two more issues were caught while re-splitting the diff (a dropped `WebAuthentication.resume()` call in the example iOS AppDelegate, and a stale Podfile deployment-target comment) and are fixed there too.

PR #829, opened 2026-05-13
feat: Adds telemetry support for flutter windows

NandanPrabhu:

### 📋 Changes

Adds the `Auth0-Client` instrumentation/telemetry header to the Windows `/oauth/token` POST request

**New — `BuildAuth0ClientHeader(name, version)` helper** (`auth0_client.h/cpp`):
- Builds a base64url-encoded JSON payload sent as the `Auth0-Client` HTTP header:
  ```json
  {"name":"auth0-flutter","version":"2.1.0","env":{"Windows":"10.0"}}
  ```
- Uses RFC 4648 §5 base64url encoding (URL-safe alphabet, no padding)
- Detects the real Windows OS version via `RtlGetVersion` (works correctly on Windows 10+ where `GetVersionEx` lies)

**Modified — `Auth0Client` class** (`auth0_client.h/cpp`):
- Constructor now accepts a third `auth0ClientHeader` string parameter
- `ExchangeCodeForTokens` attaches the `Auth0-Client` header on the token exchange request when the value is non-empty

**Modified — `LoginWebAuthRequestHandler`** (`login_web_auth_request_handler.cpp`):
- Extracts the `_userAgent` map (`name` + `version`) from Dart method call arguments
- Calls `BuildAuth0ClientHeader(name, version)` and passes the result into `Auth0Client`
- The Dart layer already sends `_userAgent` on every `WebAuthRequest` via the platform interface `Request` base class — no Dart changes needed

**Platform consistency:**

| Platform | env field |
|----------|-----------|
| iOS/macOS | `{"iOS":"17.0","swift":"5.9"}` |
| Android | `{"android":"34"}` |
| **Windows** | `{"Windows":"10.0"}` |

All platforms use the same JSON schema and base64url encoding.

### 📎 References

- Auth0.swift telemetry implementation: `Auth0/Telemetry.swift` (sets `Auth0-Client` header on every request)
- Dart-side `_userAgent` source: `auth0_flutter_platform_interface/lib/src/request/request.dart:16`
- RFC 4648 §5 (base64url encoding): https://datatracker.ietf.org/doc/html/rfc4648#section-5

### 🎯 Testing

**Automated (5 unit tests in `auth0_client_telemetry_test.cpp`):**
- `IsNotEmpty` — header is always produced
- `ContainsNoBase64Padding` — no `=` padding characters
- `ContainsNoStandardBase64Chars` — `+` → `-`, `/` → `_`
- `DecodestoValidJson` — decoded payload contains expected name, version, env, and Windows keys
- `EmptyNameAndVersionStillProducesValidHeader` — graceful fallback when Dart doesn't supply agent info

**Manual verification:**
1. Run the example app on Windows, complete a login flow
2. Check Auth0 tenant logs to confirm the `Auth0-Client` header is present on the `/oauth/token` request
3. Decode the header value (base64url → JSON) and verify it contains the correct SDK name, version, and Windows OS version


auth0_flutter/windows/auth0_client.cpp
```diff
@@ -11,16 +16,84 @@ using namespace web::http;
...
+
+    std::ostringstream oss;
+    oss << osvi.dwMajorVersion << "." << osvi.dwMinorVersion;
+    return oss.str();
+}
+
+std::string BuildAuth0ClientHeader(const std::string &name, const std::string &version)
+{
+    // { "name": "auth0-flutter", "version": "x.y.z", "env": { "Windows": "10.0" } }
+    std::ostringstream json;
```
sanchitmehtagit: String concatenation here can be a bit fragile and may lead to malformed JSON if `name` or `version` contain quotes or backslashes. Even though this is coming from the SDK and shouldn’t cause issues in practice, it would be good to clean this up for safety and consistency.
We’re already using `web::json::value` in `ExchangeCodeForTokens`, so we could follow the same approach here to ensure special characters are properly escaped.

sanchitmehtagit (changes requested):

auth0_flutter/windows/test/auth0_api_client_telemetry_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+    return result;
+}
+
+TEST(BuildAuth0ClientHeaderTest, IsNotEmpty)
+{
+    auto header = BuildAuth0ClientHeader("auth0-flutter", "2.1.0");
+    EXPECT_FALSE(header.empty());
+}
+
+TEST(BuildAuth0ClientHeaderTest, ContainsNoBase64Padding)
```
sanchitmehtagit, line 54: Thanks for adding the test for BuildAuth0ClientHeader 🙌
Would you be open to also adding a test to verify that Auth0Client actually uses the header when making the request? It would help round out the coverage nicely.

sanchitmehtagit (dismissed):

sanchitmehtagit (dismissed):
I havent tested it , LGTM otherwise

auth0_flutter/windows/test/auth0_api_client_telemetry_test.cpp
NandanPrabhu: done added it

auth0_flutter/windows/auth0_client.cpp
NandanPrabhu: good catch. Updated it

sanchitmehtagit (approved):

PR #835, opened 2026-05-20
feat: add My Account API support for managing MFA authentication method

utkrishtsahu:

### 📋 Changes

Adds support for the [Auth0 My Account API](https://auth0.com/docs/manage-users/my-account-api), enabling end-users to self-manage their MFA authentication methods without requiring Management API tokens.

**New public API:**

- Auth0.myAccount(accessToken:) — factory method returning a MyAccountApi instance
- MyAccountApi class with the following methods:
- getAuthenticationMethods() — list all enrolled authentication methods
- getAuthenticationMethod(id:) — get a specific method by ID
- deleteAuthenticationMethod(id:) — delete a method by ID
- getFactors() — list available factors on the tenant
- enrollPhone(phoneNumber:, type:) — enroll phone (SMS/voice)
- enrollEmail(email:) — enroll email OTP
- enrollTotp() — enroll TOTP (authenticator app)
- enrollPush() — enroll push notifications
- enrollRecoveryCode() — enroll recovery codes
- verifyOtp(id:, authSession:, otp:) — confirm enrollment with OTP

**New types added:**

- MyAccountApi — main API class
- AuthenticationMethod — represents an enrolled MFA method
- EnrollmentChallenge — returned by enrollment methods (contains id, authSession, and type-specific fields like totpSecret, totpUri, barcodeUri, recoveryCode)
- Factor — represents an available factor (name, enabled)
- PhoneType — enum (sms, voice)
- MyAccountException — domain exception with statusCode, isNetworkError, isRetryable

**Platform implementation:**

- Android: Auth0FlutterMyAccountMethodCallHandler + 10 individual request handlers via MyAccountAPIClient
- iOS/macOS: MyAccountHandler + 10 individual method handlers via Auth0.swift MyAccount client
- Method channel: auth0.com/auth0_flutter/my_account

Requires access tokens with audience https://{domain}/me/ and appropriate me scopes (read:me:authentication_methods, create:me:authentication_methods, delete:me:authentication_methods, read:me:factors).

### 📎 References
SDK-8730

### 🎯 Testing

**Unit tests:**

- Android: 11 new test files covering all request handlers + method call handler routing (247 tests total, all passing)
- iOS: 39 new tests covering all method handlers via spy/mock objects (250 tests total, all passing)

**Manual testing:**

- Tested on iOS Simulator (iPhone 16, iOS 18.4) and Android emulator
- Verified full enrollment flow: login with My Account audience → enroll phone (SMS) → verify OTP → list methods → delete method
- Verified TOTP enrollment returns correct totpSecret and totpUri for authenticator app integration

**Not tested:**

- Web platform (not supported )

## Summary by CodeRabbit

* **New Features**
  * Mobile My Account API (Android/iOS): list/get/delete authentication methods and factors; enroll/manage phone (SMS/voice), email, TOTP, push, recovery codes; verify OTP and confirm enrollments.
  * Auth0.myAccount(...) mobile API and example MyAccountCard widget demonstrating full flows.
  * Optional DPoP support for My Account requests.

* **Documentation**
  * Added mobile-only My Account API reference with usage, DPoP notes, and error-handling examples.


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/DeleteAuthenticationMethodRequestHandler.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.auth0.android.myaccount.MyAccountException
+import com.auth0.auth0_flutter.request_handlers.MethodCallRequest
+import com.auth0.auth0_flutter.toMyAccountMap
+import com.auth0.auth0_flutter.utils.assertHasProperties
+import io.flutter.plugin.common.MethodChannel
+
+private const val MY_ACCOUNT_DELETE_AUTH_METHOD_METHOD =
+    "myAccount#deleteAuthenticationMethod"
+
+class DeleteAuthenticationMethodRequestHandler :
```
NandanPrabhu, line 14: is final class required here?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/VerifyOtpRequestHandler.kt
```diff
@@ -0,0 +1,51 @@
...
+                        exception.getCode(),
+                        exception.getDescription(),
+                        exception.toMyAccountMap()
+                    )
+                }
+
+                override fun onSuccess(
+                    res: AuthenticationMethod
+                ) {
+                    result.success(null)
```
NandanPrabhu: is flutter compatible with suspend functions?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetFactorsRequestHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+                        exception.getDescription(),
+                        exception.toMyAccountMap()
+                    )
+                }
+
+                override fun onSuccess(res: List<Factor>) {
+                    result.success(res.map {
+                        mapOf(
+                            "name" to it.type,
+                            "enabled" to true
```
NandanPrabhu: Factor has type and usage. Dont we want to pass usage as well to dart?

sanchitmehtagit: "enabled" : true should this be derived from response.?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetFactorsRequestHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+                        exception.getCode(),
+                        exception.getDescription(),
+                        exception.toMyAccountMap()
+                    )
+                }
+
+                override fun onSuccess(res: List<Factor>) {
+                    result.success(res.map {
+                        mapOf(
+                            "name" to it.type,
```
pmathew92: Lets keep it as type only. Its not actually a  name  thats returned.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetFactorsRequestHandler.kt
pmathew92: Agree, usage array should be returned

pmathew92: What exactly does this `enabled` represent. The API doesn't return an enabled property

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetAuthenticationMethodsRequestHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+
+class GetAuthenticationMethodsRequestHandler : MyAccountRequestHandler {
+    override val method: String = MY_ACCOUNT_GET_AUTH_METHODS_METHOD
+
+    override fun handle(
+        client: MyAccountAPIClient,
+        request: MethodCallRequest,
+        result: MethodChannel.Result
+    ) {
+        client.getAuthenticationMethods()
```
pmathew92: This API now supports a `type` parameter as input. Android PR is already up.
https://github.com/auth0/Auth0.Android/pull/974
Both platforms will have this out by next week. Update accordingly

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/VerifyOtpRequestHandler.kt
pmathew92: Why is the success returning null here. Shouldn't the details from `AuthenticationMethod` type be returned here ?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterMyAccountMethodCallHandler.kt
```diff
@@ -0,0 +1,30 @@
...
+) : MethodCallHandler {
+    lateinit var context: Context
+
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+        val request = MethodCallRequest.fromCall(call)
+
+        val handler = myAccountRequestHandlers.find { it.method == call.method }
+        if (handler != null) {
+            val accessToken = request.data["accessToken"] as? String ?: ""
+            val client = MyAccountAPIClient(request.account, accessToken)
```
pmathew92: MyAccountAPI's support DPoP too. Ensure the client created can support DPoP also. 
https://github.com/auth0/Auth0.Android/pull/974
This PR adds the support for the same in Android

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExceptionExtensions.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.auth0.android.myaccount.MyAccountException
+
+fun MyAccountException.toMyAccountMap(): Map<String, Any> {
+    val exception = this
+    return buildMap {
+        put("_statusCode", exception.statusCode)
+        put("_errorFlags", mapOf(
+            "isNetworkError" to exception.isNetworkError,
+        ))
```
pmathew92, line 13: The `MyAccountException` returns more info which actually represents what went wrong like title, detail etc. These help the customer know what went wrong and also help us debug when an issue arises. Currently the above map captures only the status code, which in itself is not helpful. Add the other properties too

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.auth0.android.result.PushNotificationAuthenticationMethod
+import com.auth0.android.result.RecoveryCodeEnrollmentChallenge
+import com.auth0.android.result.TotpAuthenticationMethod
+import com.auth0.android.result.TotpEnrollmentChallenge
+
+fun AuthenticationMethod.toMyAccountMethodMap(): Map<String, Any?> {
+    return buildMap {
+        put("id", id)
+        put("type", type)
+        put("created_at", createdAt)
```
pmathew92, line 18: there is a `usage ` property also

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+                put("name", method.name)
+                put("email", method.email)
+            }
+            is TotpAuthenticationMethod -> {
+                put("name", method.name)
+            }
+            is PushNotificationAuthenticationMethod -> {
+                put("name", method.name)
+            }
+            else -> {}
```
pmathew92: Passkey , Recovery code and Password are missing

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+            }
+            is EmailAuthenticationMethod -> {
+                put("name", method.name)
+                put("email", method.email)
+            }
+            is TotpAuthenticationMethod -> {
+                put("name", method.name)
+            }
+            is PushNotificationAuthenticationMethod -> {
+                put("name", method.name)
```
pmathew92, line 34: The `confirmed` property is also missing which is important to know whether a authentication method is verfied or nor

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+            is TotpEnrollmentChallenge -> {
+                put("barcode_uri", challenge.barcodeUri)
+                put("totp_secret", challenge.manualInputCode)
+                put("totp_uri", challenge.manualInputCode)
+            }
+            is RecoveryCodeEnrollmentChallenge -> {
+                put("recovery_code", challenge.recoveryCode)
+            }
+            is MfaEnrollmentChallenge -> {}
+            else -> {}
```
pmathew92, line 61: Oob enrollment challenge is missing

pmathew92 (changes requested):

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/DeleteAuthenticationMethodRequestHandler.kt
utkrishtsahu: No, final class is not used here — it's a regular class implementing the MyAccountRequestHandler interface. Since it's implementing an interface, it can't be final. No change needed.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/VerifyOtpRequestHandler.kt
utkrishtsahu: Flutter's MethodChannel on Android uses callback-based MethodChannel.Result, not coroutines. The Auth0 Android SDK's Request.start(Callback) pattern is the correct approach for MethodChannel handlers. Suspend functions would require a coroutine scope which adds unnecessary complexity here. This is the same pattern used by all other handlers in the plugin (WebAuth, Auth, CredentialsManager).

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetFactorsRequestHandler.kt
utkrishtsahu: Fixed. Updated to pass "type" and "usage" directly from the native Factor object instead of mapping to "name"/"enabled".

utkrishtsahu: Fixed. Removed enabled entirely — the API doesn't return an enabled property. Replaced with "usage" which is the actual field from the SDK (List<String>? like ["secondary"]).

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetFactorsRequestHandler.kt
utkrishtsahu: Fixed. Changed from "name" to it.type to "type" to it.type.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetFactorsRequestHandler.kt
utkrishtsahu: You're right — removed it. Replaced with the actual usage array from the SDK. Updated the Dart Factor model accordingly (type + usage instead of name + enabled).

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/GetAuthenticationMethodsRequestHandler.kt
utkrishtsahu: Acknowledged. Will add the optional type filter parameter to getAuthenticationMethods() once Auth0.Android PR #974 is merged and the corresponding Swift SDK update is available. Will address in a follow-up commit/PR.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/my_account/VerifyOtpRequestHandler.kt
utkrishtsahu: Fixed. Now returning res.toMyAccountMethodMap() on success. Updated the Dart return type from Future<void> to Future<AuthenticationMethod> across the platform interface and public API. iOS handler also updated to return method.asDictionary().

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterMyAccountMethodCallHandler.kt
utkrishtsahu: Acknowledged. Will add DPoP support to the MyAccountAPIClient creation once Auth0.Android PR #974 is merged and the Swift SDK adds equivalent support. This will involve passing a useDPoP flag from Dart and configuring the client accordingly. Will address in a follow-up once both native SDKs have this merged.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExceptionExtensions.kt
utkrishtsahu: Fixed. Added title and detail to the error map on both Android and iOS. Also exposed them as first-class properties on the Dart MyAccountException class (exception.title, exception.detail).

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
utkrishtsahu: Fixed. Added usage to the AuthenticationMethod serialization (inside the MfaAuthenticationMethod check alongside confirmed). Also added it to the Dart model.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
utkrishtsahu: Added an else branch that sets name to null for unhandled subtypes (Passkey, RecoveryCode, Password). These types don't have additional properties beyond the base id, type, createdAt that are already serialized for all AuthenticationMethod instances. Full passkey enrollment support is deferred to a future release due to WebAuthn/FIDO2 platform complexity.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
utkrishtsahu: Fixed. confirmed was already being serialized on Android (inside the MfaAuthenticationMethod check at line 39). Also added it to the iOS AuthenticationMethod.asDictionary() and the Dart AuthenticationMethod model so it's accessible as method.confirmed.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
utkrishtsahu: Acknowledged. Will add OOB (Out-of-Band) enrollment challenge support in a follow-up once the full OOB enrollment flow is confirmed across both native SDKs.

auth0_flutter_platform_interface/lib/src/myaccount/phone_type.dart
```diff
@@ -0,0 +1,24 @@
...
+  static PhoneType fromValue(final String value) {
+    switch (value) {
+      case 'sms':
+        return PhoneType.sms;
+      case 'voice':
+        return PhoneType.voice;
+      default:
+        return PhoneType.sms;
+    }
+  }
```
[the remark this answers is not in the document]
utkrishtsahu: This is on the existing PhoneType.fromValue. The new AuthenticationMethodType.fromValue already returns null for unknown values, which is the safe pattern for deserialization. Keeping PhoneType unchanged to avoid an unrelated behavior change.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountExtensions.swift
```diff
@@ -0,0 +1,107 @@
...
+            "email": nil,
+            "totp_secret": nil,
+            "totp_uri": nil,
+            "preferred_authentication_method": preferredAuthenticationMethod,
+            "created_at": createdAt,
+            "last_auth_at": nil,
+            "confirmed": confirmed,
+            "usage": usage
+        ]
+    }
```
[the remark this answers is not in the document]
utkrishtsahu: In Auth0.swift 2.21.0 the AuthenticationMethod struct declares createdAt as String (only PasskeyAuthenticationMethod uses Date). So "created_at": createdAt already passes a codec-safe String and the Dart side parses it with DateTime.parse. No change needed.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountVerifyOtpMethodHandler.swift
```diff
@@ -0,0 +1,35 @@
...
+        guard let authSession = arguments["authSession"] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing("authSession")))
+        }
+        guard let otp = arguments["otp"] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing("otp")))
+        }
+
+        client
+            .authenticationMethods
+            .confirmPhoneEnrollment(id: id, authSession: authSession, otpCode: otp)
```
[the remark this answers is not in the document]
utkrishtsahu: This is the existing OTP-confirm design, not part of this change. All confirm-with-OTP factors hit the same POST {id, authSession, otp_code} endpoint, so the method name is interchangeable on the wire. Out of scope here.

auth0_flutter/example/ios/Tests/MyAccount/MyAccountConfirmEnrollmentMethodHandlerTests.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    func testProducesFlutterErrorOnFailure() {
+        let expectation = self.expectation(description: "Produced FlutterError")
+        spy.confirmResult = .failure(MyAccountError(info: [:], statusCode: 403))
+        sut.handle(with: ["id": "test-id", "authSession": "session"]) { result in
+            XCTAssertTrue(result is FlutterError)
+            expectation.fulfill()
+        }
+        wait(for: [expectation])
+    }
```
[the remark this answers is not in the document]
utkrishtsahu: There's a custom wait(for:) extension in Tests/Utilities.swift that supplies a default timeout. All existing and new MyAccount tests use it, and they compile/run in CI.

auth0_flutter/lib/auth0_flutter.dart
```diff
@@ -3,29 +3,36 @@ import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interfac
...
-        Credentials,
         SSOCredentials,
         UserProfile,
-        DatabaseUser,
-        ChallengeType,
-        CredentialsManagerException,
-        PasswordlessType,
-        LocalAuthentication,
-        LocalAuthenticationLevel;
+        WebAuthenticationException;
```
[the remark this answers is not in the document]
utkrishtsahu: Good catch — fixed. Added AuthenticationMethodType to the public show list in auth0_flutter.dart so the type: filter is usable by consumers.

auth0_flutter/README.md
```diff
@@ -632,6 +632,101 @@ void dispose() {
...
+  audience: 'https://YOUR_DOMAIN/me/',
+  scopes: {
+    'openid',
+    'read:me:authentication_methods',
+    'create:me:authentication_methods',
+    'update:me:authentication_methods',
+    'delete:me:authentication_methods',
+    'read:me:enrollments',
+    'read:me:factors',
+  },
```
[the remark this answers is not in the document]
utkrishtsahu: Already addressed — the scopes block includes both read:me:enrollments and read:me:factors.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
```diff
@@ -0,0 +1,64 @@
...
+                put("name", method.name)
+            }
+            is PushNotificationAuthenticationMethod -> {
+                put("name", method.name)
+            }
+            else -> {
+                put("name", null)
+            }
+        }
+        if (this@toMyAccountMethodMap is MfaAuthenticationMethod) {
```
pmathew92, line 41: This will add `usage` for only MfaAuthenticationMethod type. Will miss it for Passkey and Password types

auth0_flutter/darwin/auth0_flutter.podspec
```diff
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.18.0'
+  s.dependency 'Auth0', '2.21.0'
```
pmathew92: Use `2.21.2` as it fixes the dpop not added to the update method

auth0_flutter/ios/auth0_flutter.podspec
```diff
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.18.0'
+  s.dependency 'Auth0', '2.21.0'
```
pmathew92: Use `2.21.2`

auth0_flutter/lib/auth0_flutter.dart
```diff
@@ -127,6 +134,33 @@ class Auth0 {
   WindowsWebAuthentication windowsWebAuthentication() =>
       WindowsWebAuthentication(_account, _userAgent);
 
+  /// Creates an instance of [MyAccountApi] for managing the user's
+  /// authentication methods (MFA factors) via the
+  /// [My Account API](https://auth0.com/docs/manage-users/my-account-api).
+  ///
+  /// Requires an [accessToken] with the appropriate `me` scopes and
```
pmathew92:
```suggestion
  /// Requires an [accessToken] with the appropriate  scopes and
```

auth0_flutter/macos/auth0_flutter.podspec
```diff
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.18.0'
+  s.dependency 'Auth0', '2.21.0'
```
pmathew92: use `2.21.0`

auth0_flutter/README.md
```diff
@@ -632,6 +632,101 @@ void dispose() {
 - [clearCredentials](https://pub.dev/documentation/auth0_flutter/latest/auth0_flutter/DefaultCredentialsManager/clearCredentials.html)
 - [ssoCredentials](https://pub.dev/documentation/auth0_flutter/latest/auth0_flutter/DefaultCredentialsManager/ssoCredentials.html)
 
+#### My Account API
+
+The My Account API allows users to manage their own multi-factor authentication (MFA) methods. Available on **mobile (Android/iOS) only**.
+
+First, authenticate with the `https://{domain}/me/` audience:
+
```
pmathew92, line 642: Are you able to authenticate with the `https://YOUR_DOMAIN/me/` audience ? Ideally the recommended approach is use MRRT to get the token for my account audience and not get them while authenticating. check how RN has done
https://github.com/auth0/react-native-auth0/blob/master/EXAMPLES.md#my-account-api

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
```diff
@@ -0,0 +1,64 @@
...
+
+fun EnrollmentChallenge.toMyAccountChallengeMap(): Map<String, Any?> {
+    return buildMap {
+        put("id", id)
+        put("auth_session", authSession)
+        when (val challenge = this@toMyAccountChallengeMap) {
+            is TotpEnrollmentChallenge -> {
+                put("barcode_uri", challenge.barcodeUri)
+                put("totp_secret", challenge.manualInputCode)
+                put("totp_uri", challenge.barcodeUri)
```
sanchitmehtagit, line 55: is this expected to set barcodeUri  for totp_uri

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterMyAccountMethodCallHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+    private val myAccountRequestHandlers: List<MyAccountRequestHandler>
+) : MethodCallHandler {
+    lateinit var context: Context
+
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
+        val request = MethodCallRequest.fromCall(call)
+
+        val handler = myAccountRequestHandlers.find { it.method == call.method }
+        if (handler != null) {
+            val accessToken = request.data["accessToken"] as? String ?: ""
```
sanchitmehtagit: should we throw IllegalArgumentException like the other handlers do via assertHasProperties if access token is missing

auth0_flutter_platform_interface/lib/src/myaccount/authentication_method.dart
```diff
@@ -0,0 +1,67 @@
+class AuthenticationMethod {
+  final String id;
+  final String type;
+  final String? name;
+  final String? phoneNumber;
+  final String? email;
+  final String? totpSecret;
+  final String? totpUri;
+  final String? preferredAuthenticationMethod;
```
sanchitmehtagit, line 9: could we use enum Phonetype here  for preferredAuthenticationMethod

auth0_flutter_platform_interface/lib/src/myaccount/authentication_method.dart
```diff
@@ -0,0 +1,67 @@
...
+        type: result['type'] as String,
+        name: result['name'] as String?,
+        phoneNumber: result['phone_number'] as String?,
+        email: result['email'] as String?,
+        totpSecret: result['totp_secret'] as String?,
+        totpUri: result['totp_uri'] as String?,
+        preferredAuthenticationMethod:
+            result['preferred_authentication_method'] as String?,
+        createdAt: result['created_at'] != null
+            ? DateTime.parse(result['created_at'] as String)
```
sanchitmehtagit: DateTime.parse(result['created_at'] as String)
This may generate type error if created_at is not string

what we could do instead here is 
```
  createdAt: result['created_at'] != null                                                       
      ? DateTime.parse(result['created_at'].toString())  // safe regardless of int/String                                              
      : null,     
```

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountExtensions.swift
```diff
@@ -0,0 +1,107 @@
...
+extension AuthenticationMethod {
+    func asDictionary() -> [String: Any?] {
+        return [
+            "id": id,
+            "type": type,
+            "name": name,
+            "phone_number": phoneNumber,
+            "email": nil,
+            "totp_secret": nil,
+            "totp_uri": nil,
```
sanchitmehtagit, line 34: is email , toto_secret and totp_uri expected to be sent as nil here

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountExtensions.swift
```diff
@@ -0,0 +1,107 @@
...
+#endif
+
+extension FlutterError {
+    convenience init(from error: MyAccountError) {
+        let details: [String: Any] = [
+            "_statusCode": error.statusCode,
+            "_title": error.title,
+            "_detail": error.detail,
+            "_errorFlags": [
+                "isNetworkError": false
```
sanchitmehtagit: isNetworkError is set as false. lets fix this

sanchitmehtagit: "_errorFlags": [                
      "isNetworkError": error.isNetworkError                                                              
  ]

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountVerifyOtpMethodHandler.swift
sanchitmehtagit: The Auth0.swift SDK exposes three separate confirm methods:                                                                          
  - confirmPhoneEnrollment(id:authSession:otpCode:)                                                                                    
  - confirmEmailEnrollment(id:authSession:otpCode:)                                                                                    
  - confirmTOTPEnrollment(id:authSession:otpCode:)                                              
                                                                                                                                       
  These hit different API endpoints. Calling confirmPhoneEnrollment with a TOTP or email enrollment ID will result in a server error,  
  meaning TOTP and email OTP enrollment can never be completed on iOS.

sanchitmehtagit (changes requested):

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountConfirmEnrollmentMethodHandler.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    func handle(with arguments: [String: Any], callback: @escaping FlutterResult) {
+        guard let id = arguments["id"] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing("id")))
+        }
+        guard let authSession = arguments["authSession"] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing("authSession")))
+        }
+
+        client
```
sanchitmehtagit: The Auth0.swift SDK exposes two separate methods for these:                                                                          
  - confirmPushNotificationEnrollment(id:authSession:)                                          
  - confirmRecoveryCodeEnrollment(id:authSession:)                                                                                     
                                                                                                
  These hit different API endpoints. Calling confirmRecoveryCodeEnrollment with a push enrollment ID will result in a server error,    
  meaning push notification enrollment can never be completed on iOS.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
utkrishtsahu: Good catch. Moved usage serialization out of the MfaAuthenticationMethod block so it's now serialized for every AuthenticationMethod subtype (confirmed via javap that usage lives on the base class). confirmed stays MFA-specific since it's only declared on MfaAuthenticationMethod.

auth0_flutter/darwin/auth0_flutter.podspec
utkrishtsahu: Done — bumped to 2.21.2.

auth0_flutter/ios/auth0_flutter.podspec
utkrishtsahu: Done

auth0_flutter/lib/auth0_flutter.dart
utkrishtsahu: done

auth0_flutter/macos/auth0_flutter.podspec
utkrishtsahu: done

auth0_flutter/README.md
utkrishtsahu: Thanks, good point. The example authenticates against the me audience directly just to keep the sample minimal. I'll add a note to the README clarifying that the recommended production approach is to obtain a My Account–audience token via MRRT (Multi-Resource Refresh Token) rather than during the initial login, mirroring how react-native-auth0 documents it.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MyAccountExtensions.kt
utkrishtsahu: Yes, intentional. The Android SDK exposes the TOTP QR-code URL as barcodeUri, while iOS exposes it as totpUri. We normalize both to the totp_uri key so the Dart model is consistent across platforms.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterMyAccountMethodCallHandler.kt
utkrishtsahu: Agreed — added assertHasProperties(listOf("accessToken"), request.data) before the cast, consistent with the other handlers.

auth0_flutter_platform_interface/lib/src/myaccount/authentication_method.dart
utkrishtsahu: I kept this as String? deliberately. It's a response field reflecting the raw API value, and typing it to PhoneType would silently drop any value the API adds in the future. Keeping it a string passes everything through transparently. Happy to switch if you feel strongly about the typed version.

auth0_flutter_platform_interface/lib/src/myaccount/authentication_method.dart
utkrishtsahu: Fixed exactly as suggested — applied .toString() to both created_at and last_auth_at.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountExtensions.swift
utkrishtsahu: Yes. The Auth0.swift AuthenticationMethod only populates the fields relevant to that method type — e.g. a phone enrollment has no totpSecret/email, so they come back nil and are simply omitted. The Dart model declares them nullable, so this is handled.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountExtensions.swift
utkrishtsahu: Fixed — changed to error.isNetworkError. MyAccountError conforms to Auth0APIError, which provides isNetworkError.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountVerifyOtpMethodHandler.swift
utkrishtsahu: Great catch — this was a real functional bug. Now routing by factorType: email → confirmEmailEnrollment, totp → confirmTOTPEnrollment, otherwise confirmPhoneEnrollment. factorType is threaded from Dart via MyAccountVerifyOtpOptions, and I added iOS handler tests asserting each factor routes to the correct SDK method.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountConfirmEnrollmentMethodHandler.swift
utkrishtsahu: Fixed — same root cause. Now routing by factorType: push-notification → confirmPushNotificationEnrollment, otherwise confirmRecoveryCodeEnrollment, threaded from Dart via MyAccountConfirmEnrollmentOptions. Added routing tests on the iOS handler to lock this in.

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountConfirmEnrollmentMethodHandler.swift
```diff
@@ -0,0 +1,40 @@
...
+        guard let factorType = arguments["factorType"] as? String else {
+            return callback(FlutterError(from: .requiredArgumentMissing("factorType")))
+        }
+
+        let request: any Request<AuthenticationMethod, MyAccountError>
+        if factorType == "push-notification" {
+            request = client.authenticationMethods.confirmPushNotificationEnrollment(id: id, authSession: authSession)
+        } else {
+            request = client.authenticationMethods.confirmRecoveryCodeEnrollment(id: id, authSession: authSession)
+        }
```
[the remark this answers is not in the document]
utkrishtsahu: done

auth0_flutter/darwin/Classes/MyAccountAPI/MyAccountVerifyOtpMethodHandler.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        let request: any Request<AuthenticationMethod, MyAccountError>
+        switch factorType {
+        case "email":
+            request = client.authenticationMethods.confirmEmailEnrollment(id: id, authSession: authSession, otpCode: otp)
+        case "totp":
+            request = client.authenticationMethods.confirmTOTPEnrollment(id: id, authSession: authSession, otpCode: otp)
+        default:
+            request = client.authenticationMethods.confirmPhoneEnrollment(id: id, authSession: authSession, otpCode: otp)
+        }
```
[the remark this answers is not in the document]
utkrishtsahu: done

sanchitmehtagit (dismissed):

auth0_flutter/README.md
pmathew92: @utkrishtsahu  I don't see this note added. I owuld still recommend to follow the proper appraoch in the examples file

pmathew92 (changes requested):

auth0_flutter/EXAMPLES.md
```diff
@@ -1289,3 +1296,139 @@ try {
 ---
 
 [Go up ⤴](#examples)
+
+## 📱 My Account API
+
+The My Account API lets authenticated users manage their own multi-factor authentication (MFA) methods — enrolling, confirming, listing, updating, and deleting factors such as phone, email, TOTP, push notifications, and recovery codes. It is available on **mobile (Android/iOS) only**.
+
+> 💡 The My Account API is currently in [Early Access](https://auth0.com/docs/troubleshoot/product-lifecycle/product-release-stages#early-access). It requires a [custom domain](https://auth0.com/docs/customize/custom-domains) and must be enabled for your tenant — reach out to Auth0 support to get it enabled.
```
pmathew92: This has been pushed to GA now. Remove the EA tag

pmathew92: MyAccount doesn't require custom domain. Passkeys require custom domain. Mention this in the passkey enroll flow

auth0_flutter/EXAMPLES.md
```diff
@@ -1289,3 +1296,139 @@ try {
...
+```dart
+// 1. Log in once for your app, requesting offline_access to get a refresh token.
+final credentials = await auth0.webAuthentication().login(
+  scopes: {'openid', 'profile', 'email', 'offline_access'},
+);
+await auth0.credentialsManager.storeCredentials(credentials);
+
+// 2. Exchange the stored refresh token for a token scoped to the My Account API,
+//    by requesting the `https://YOUR_DOMAIN/me/` audience and the My Account scopes.
+final myAccountCredentials = await auth0.credentialsManager.credentials(
```
pmathew92: the problem with this approach is that this will replace the existing Credentials with the MyAccount one and user won't be able to use the first AT anymore.  We would need to support MRRT for the same

pmathew92 (approved):

PR #836, opened 2026-05-20
fix: use safe casts in UserProfileExtensions to prevent ClassCastException

### LINKED ISSUE #830: `ClassCastException: LinkedTreeMap cannot be cast to String` when ID token claims contain JSON objects instead of strings
### Checklist

- [x] The issue can be reproduced in the [auth0_flutter sample app](https://github.com/auth0-samples/auth0-flutter-samples/tree/main/sample) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/auth0-flutter/tree/main/auth0_flutter#readme), [Examples](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/EXAMPLES.md), and [FAQ](https://github.com/auth0/auth0-flutter/blob/main/auth0_flutter/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://pub.dev/documentation/auth0_flutter/latest/) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/auth0-flutter/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

## Description

When a user authenticates via a social provider that returns standard OIDC claims (such as `picture` or `profile`) as JSON objects rather than plain strings, the Android SDK crashes with a fatal `ClassCastException` immediately after login.

## Steps to Reproduce

1. Set up Auth0 with a social connection where the user profile stores `picture` (or `profile`) as a JSON object, e.g. `{"url": "https://...", "type": "image/jpeg"}` rather than a plain string URL
2. Call `auth0.webAuthentication(scheme: '...').login()` on Android
3. App crashes immediately after the browser closes

## Error

```
FATAL EXCEPTION: main
java.lang.ClassCastException: com.google.gson.internal.LinkedTreeMap cannot be cast to java.lang.String
    at com.auth0.auth0_flutter.UserProfileExtensionsKt.getProfileURL(UserProfileExtensions.kt:42)
    at com.auth0.auth0_flutter.UserProfileExtensionsKt.toMap(UserProfileExtensions.kt:15)
    at com.auth0.auth0_flutter.request_handlers.web_auth.LoginWebAuthRequestHandler$handle$1.onSuccess(LoginWebAuthRequestHandler.kt:95)
```

## Root Cause

`UserProfileExtensions.kt` uses unsafe Kotlin casts (`as String?`) for all standard OIDC string claims. When GSON parses a JSON object value it produces a `LinkedTreeMap`, and `as String?` throws `ClassCastException` on any non-null non-String value (unlike `as? String` which would safely return `null`).

Affected properties: `profileURL`, `websiteURL`, `gender`, `birthdate`, `zoneinfo`, `locale`, `phoneNumber`, `updatedAt`, `middleName`, `preferredUsername` — and `pictureURL` via the underlying `auth0-android` `UserProfile` class.

## Expected Behaviour

Claims that are unexpectedly non-string values should be treated as `null` rather than crashing the app.

## Suggested Fix

Change all `as String?` (unsafe nullable cast) to `as? String` (safe cast) in `UserProfileExtensions.kt`, and read `picture` directly from `getExtraInfo()` to bypass the base class property:

```kotlin
fun UserProfile.toMap(): Map<String, Any?> {
    return mapOf(
        ...
        "picture" to (this.getExtraInfo()["picture"] as? String),
        ...
    )
}

val UserProfile.profileURL: String?
    get() = getExtraInfo()["profile"] as? String

val UserProfile.middleName: String?
    get() = getExtraInfo()["middle_name"] as? String

// ... same pattern for all other String? extension properties
```

## Environment

- `auth0_flutter`: 2.1.0-beta.1
- Android (Samsung Galaxy S21)
- Social connection returning `picture` / `profile` as JSON object

---

File it at: **https://github.com/auth0/auth0-flutter/issues/new**

### Reproduction

1. Set up Auth0 with a social connection where the user profile stores `picture` or `profile` as a JSON object (e.g. `{"url": "https://..."}`) rather than a plain string URL
2. Call `auth0.webAuthentication(scheme: '...').login()` on Android
3. App crashes immediately after the browser closes and the SDK tries to build the UserProfile

### Additional context

The crash also occurs on the `profile` claim, not just `picture`. Any standard OIDC claim that a social provider returns as a JSON object instead of a string will trigger this. The underlying issue is GSON deserialising JSON objects as LinkedTreeMap, which then cannot be cast with `as String?`.

### auth0_flutter version

2.1.0-beta.1

### Flutter version

3.41.9

### Platform

Android

### Platform version(s)

Android 13 (Samsung Galaxy S21)

sanchitmehtagit:
## Summary

Fixes a fatal `ClassCastException` on Android when a social provider returns standard OIDC claims (e.g. `picture`, `profile`) as JSON objects instead of plain strings.

Closes #830

## Problem

`UserProfileExtensions.kt` used unsafe Kotlin casts (`as String?`) for all string-typed OIDC claims read from `getExtraInfo()`. When GSON deserializes a JSON object value it produces a `LinkedTreeMap`. An unsafe `as String?` cast only short-circuits for `null` — it still throws `ClassCastException` for any non-null, non-String value. This caused an immediate crash after login whenever the user profile contained a claim with a non-string value.

Additionally, `picture` was read via `this.pictureURL` which delegates to the base `UserProfile` class property — also an unsafe cast and not overridable by an extension property.

### `picture` in `toMap()`
- Tries `runCatching { this.pictureURL }.getOrNull()` first — preserves the base class value (e.g. set via constructor) and catches any `ClassCastException` from the SDK.
- Falls back to `getExtraInfo()["picture"] as? String` — returns `null` instead of crashing if the value is a non-string type.

This two-step approach is necessary because:
- `pictureURL` is a base class member property — extension properties cannot override member properties in Kotlin.
- The base class property does its own unsafe cast internally, so we wrap it in `runCatching` rather than bypassing it entirely.

### All string extension properties
Replaced every `as T?` unsafe cast with `as? T` safe cast. Returns `null` instead of throwing when the value is an unexpected type.

Affected: `middleName`, `preferredUsername`, `profileURL`, `websiteURL`, `gender`, `birthdate`, `zoneinfo`, `locale`, `phoneNumber`, `isPhoneNumberVerified`, `updatedAt`.

## Testing

Added unit tests for the crash scenarios:

- `picture` as a JSON object returns `null` instead of crashing
- All string claims as non-string values return `null` instead of crashing

### DIFF
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/UserProfileExtensions.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/UserProfileExtensions.kt
@@ -13,7 +13,7 @@ fun UserProfile.toMap(): Map<String, Any?> {
         "nickname" to this.nickname,
         "preferred_username" to this.preferredUsername,
         "profile" to this.profileURL,
-        "picture" to (getExtraInfo()["picture"] as? String),
+        "picture" to (runCatching { this.pictureURL }.getOrNull() ?: getExtraInfo()["picture"] as? String),
         "website" to this.websiteURL,
         "email" to this.email,
         "email_verified" to this.isEmailVerified,
@@ -66,7 +66,7 @@ val UserProfile.updatedAt: String?
     get() = getExtraInfo()["updated_at"] as? String
 
 val UserProfile.address: Map<String, String>?
-    get() = @Suppress("UNCHECKED_CAST") (getExtraInfo()["address"] as? Map<String, String>)
+    get() = getExtraInfo()["address"] as? Map<String, String>
 
 val UserProfile.customClaims: Map<String, Any>
     get() = getCustomClaims(getExtraInfo())
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/UserProfileExtensions.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/UserProfileExtensions.kt
@@ -13,7 +13,7 @@ fun UserProfile.toMap(): Map<String, Any?> {
         "nickname" to this.nickname,
         "preferred_username" to this.preferredUsername,
         "profile" to this.profileURL,
-        "picture" to this.pictureURL,
+        "picture" to (getExtraInfo()["picture"] as? String),
         "website" to this.websiteURL,
         "email" to this.email,
         "email_verified" to this.isEmailVerified,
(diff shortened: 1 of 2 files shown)


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/UserProfileExtensions.kt
```diff
@@ -13,7 +13,7 @@ fun UserProfile.toMap(): Map<String, Any?> {
         "nickname" to this.nickname,
         "preferred_username" to this.preferredUsername,
         "profile" to this.profileURL,
-        "picture" to this.pictureURL,
+        "picture" to (runCatching { this.pictureURL }.getOrNull() ?: getExtraInfo()["picture"] as? String),
```
sanchitmehtagit, line 16: issue reporter suggested to read it directly from extraInfo but i have kept it as a fallback instead. and also have asked query here
https://github.com/auth0/auth0-flutter/issues/830#issuecomment-4507793063

sanchitmehtagit:
https://openid.net/specs/openid-connect-core-1_0.html#AddressClaim
Address claims are supposed to have key and value as string so this will be safe downstream

utkrishtsahu (approved):

pmathew92 (approved):

PR #846, opened 2026-05-26
feat: add Partial Custom Tabs support for Android

utkrishtsahu:

### 📋 Changes

Adds support for [Partial Custom Tabs](https://developer.chrome.com/docs/android/custom-tabs/guide-partial-custom-tabs) in auth0-flutter, allowing the authentication page to be displayed as a bottom sheet or side sheet on Android instead of a full-screen browser tab.

**New types:**

- CustomTabsOptions — Dart class mirroring the iOS SafariViewController pattern, with properties: initialHeight, resizable, toolbarCornerRadius, initialWidth, sideSheetBreakpoint, backgroundInteractionEnabled, allowedBrowsers

**Modified methods:**

- WebAuthentication.login() — added optional customTabsOptions parameter
- WebAuthentication.logout() — added optional customTabsOptions parameter

**Deprecated:**

- Top-level allowedBrowsers parameter on login() and logout() — replaced by CustomTabsOptions.allowedBrowsers

**Native (Kotlin):**

- LoginWebAuthRequestHandler and LogoutWebAuthRequestHandler parse the new customTabsOptions map and build CustomTabsOptions via the Auth0.Android SDK builder
- Bumped com.auth0.android:auth0 from 3.12.0 to 3.16.0

**Backward compatibility:** Fully non-breaking. The new parameter defaults to null. The deprecated allowedBrowsers continues to work as a fallback when customTabsOptions is absent.

**Usage:**

`await auth0.webAuthentication().login(
  customTabsOptions: const CustomTabsOptions(
    initialHeight: 700,
    toolbarCornerRadius: 16,
    backgroundInteractionEnabled: true,
  ),
);
`

### 📎 References

SDK-8720

### 🎯 Testing

**Unit tests added:**

- 5 Dart tests in web_authentication_test.dart — verifies customTabsOptions is serialized correctly for login/logout, defaults to null, and side sheet options pass through
- 5 Kotlin tests in LoginWebAuthRequestHandlerTest.kt — verifies partial tabs parsing with initialHeight, all options, absence, priority over top-level allowedBrowsers, and fallback behavior
- 5 Kotlin tests in LogoutWebAuthRequestHandlerTest.kt — same coverage as login handler

**Not tested:**

- No iOS changes; customTabsOptions is Android-only and ignored on other platforms

### DIFF
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
@@ -44,14 +44,56 @@ class LoginWebAuthRequestHandler(
             builder.withInvitationUrl(args["invitationUrl"] as String)
         }
 
-        val allowedBrowsers = (args["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
-        if(allowedBrowsers.isNotEmpty()) {
-            builder.withCustomTabsOptions(
-                CustomTabsOptions.newBuilder().withBrowserPicker(
+        val customTabsOptionsMap = args["customTabsOptions"] as? Map<*, *>
+        val allowedBrowsers = if (customTabsOptionsMap != null) {
+            (customTabsOptionsMap["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
+        } else {
+            (args["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
+        }
+
+        if (customTabsOptionsMap != null || allowedBrowsers.isNotEmpty()) {
+            val ctBuilder = CustomTabsOptions.newBuilder()
+
+            if (allowedBrowsers.isNotEmpty()) {
+                ctBuilder.withBrowserPicker(
                     BrowserPicker.newBuilder()
                         .withAllowedPackages(allowedBrowsers).build()
-                ).build()
-            )
+                )
+            }
+
+            if (customTabsOptionsMap != null) {
+                val initialHeight = customTabsOptionsMap["initialHeight"] as? Int
+                if (initialHeight != null && initialHeight > 0) {
+                    ctBuilder.withInitialHeight(initialHeight)
+                }
+
+                val resizable = customTabsOptionsMap["resizable"] as? Boolean
+                if (resizable != null) {
+                    ctBuilder.withResizable(resizable)
+                }
+
+                val toolbarCornerRadius = customTabsOptionsMap["toolbarCornerRadius"] as? Int
+                if (toolbarCornerRadius != null && toolbarCornerRadius > 0) {
+                    ctBuilder.withToolbarCornerRadius(toolbarCornerRadius)
+                }
+
+                val initialWidth = customTabsOptionsMap["initialWidth"] as? Int
+                if (initialWidth != null && initialWidth > 0) {
+                    ctBuilder.withInitialWidth(initialWidth)
+                }
+
+                val sideSheetBreakpoint = customTabsOptionsMap["sideSheetBreakpoint"] as? Int
+                if (sideSheetBreakpoint != null && sideSheetBreakpoint > 0) {
+                    ctBuilder.withSideSheetBreakpoint(sideSheetBreakpoint)
+                }
+
+                val backgroundInteractionEnabled = customTabsOptionsMap["backgroundInteractionEnabled"] as? Boolean
+                if (backgroundInteractionEnabled == true) {
+                    ctBuilder.withBackgroundInteractionEnabled(true)
+                }
+            }
+
+            builder.withCustomTabsOptions(ctBuilder.build())
         }
 
         if (args["leeway"] is Int) {
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
@@ -3,12 +3,11 @@ package com.auth0.auth0_flutter.request_handlers.web_auth
 import android.content.Context
 import com.auth0.android.authentication.AuthenticationException
 import com.auth0.android.callback.Callback
-import com.auth0.android.provider.BrowserPicker
-import com.auth0.android.provider.CustomTabsOptions
 import com.auth0.android.provider.WebAuthProvider
 import com.auth0.android.result.Credentials
 import com.auth0.auth0_flutter.request_handlers.MethodCallRequest
 import com.auth0.auth0_flutter.toMap
+import com.auth0.auth0_flutter.utils.buildCustomTabsOptions
 import io.flutter.plugin.common.MethodChannel
 import java.util.*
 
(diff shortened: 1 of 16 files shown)


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
```diff
@@ -44,14 +44,56 @@ class LoginWebAuthRequestHandler(
...
-        val allowedBrowsers = (args["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
-        if(allowedBrowsers.isNotEmpty()) {
-            builder.withCustomTabsOptions(
-                CustomTabsOptions.newBuilder().withBrowserPicker(
+        val customTabsOptionsMap = args["customTabsOptions"] as? Map<*, *>
+        val allowedBrowsers = if (customTabsOptionsMap != null) {
+            (customTabsOptionsMap["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
+        } else {
+            (args["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
+        }
```
pmathew92: Should we handle the edge scenario where user can send both `customTabsOptions` and `allowedBrowsers` ?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
```diff
@@ -44,14 +44,56 @@ class LoginWebAuthRequestHandler(
...
+                ctBuilder.withBrowserPicker(
                     BrowserPicker.newBuilder()
                         .withAllowedPackages(allowedBrowsers).build()
-                ).build()
-            )
+                )
+            }
+
+            if (customTabsOptionsMap != null) {
+                val initialHeight = customTabsOptionsMap["initialHeight"] as? Int
```
pmathew92: Pbly extract all these checks to a Util method to be reused for both login and logout ?

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
```diff
@@ -392,4 +392,74 @@ class LoginWebAuthRequestHandlerTest {
...
+    fun `handler should apply customTabsOptions with initialHeight`() {
+        val args = hashMapOf<String, Any?>(
+            "customTabsOptions" to mapOf(
+                "initialHeight" to 700,
+                "allowedBrowsers" to listOf<String>()
+            )
+        )
+
+        runRequestHandler(args) { _, builder ->
+            verify(builder).withCustomTabsOptions(any())
```
pmathew92, line 403: This is validating any property being passed. Not the exact property as mentioned in the title

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
```diff
@@ -392,4 +392,74 @@ class LoginWebAuthRequestHandlerTest {
...
+                "toolbarCornerRadius" to 16,
+                "initialWidth" to 500,
+                "sideSheetBreakpoint" to 840,
+                "backgroundInteractionEnabled" to true,
+                "allowedBrowsers" to listOf("com.android.chrome")
+            )
+        )
+
+        runRequestHandler(args) { _, builder ->
+            verify(builder).withCustomTabsOptions(any())
```
pmathew92, line 422: Same as above

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
```diff
@@ -392,4 +392,74 @@ class LoginWebAuthRequestHandlerTest {
...
+        val args = hashMapOf<String, Any?>(
+            "allowedBrowsers" to listOf("org.mozilla.firefox"),
+            "customTabsOptions" to mapOf(
+                "initialHeight" to 700,
+                "allowedBrowsers" to listOf("com.android.chrome")
+            )
+        )
+
+        runRequestHandler(args) { _, builder ->
+            verify(builder).withCustomTabsOptions(any())
```
pmathew92, line 448: Validation is vague here ?

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
```diff
@@ -392,4 +392,74 @@ class LoginWebAuthRequestHandlerTest {
...
+    }
+
+    @Test
+    fun `handler should fall back to top-level allowedBrowsers when customTabsOptions is absent`() {
+        val args = hashMapOf<String, Any?>(
+            "allowedBrowsers" to listOf("com.android.chrome", "org.mozilla.firefox")
+        )
+
+        runRequestHandler(args) { _, builder ->
+            verify(builder).withCustomTabsOptions(any())
```
pmathew92, line 459: Same as above

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
utkrishtsahu: Yes, this is already handled! When both are provided, customTabsOptions.allowedBrowsers wins and the top-level one is ignored. I've now extracted this into a shared utility with a KDoc comment that makes the precedence explicit. There's also a test specifically for this case (uses allowedBrowsers from customTabsOptions when both are provided).

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
utkrishtsahu: Created buildCustomTabsOptions() in the utils package. Both handlers now just do:


buildCustomTabsOptions(args)?.let { builder.withCustomTabsOptions(it) }
Much cleaner and keeps them in sync.

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
utkrishtsahu: Fair point. The issue is that CustomTabsOptions has all private fields with no public getters, so we can't directly assert on the built object from the handler level. I've added a dedicated CustomTabsOptionsBuilderTest that tests our utility function and uses reflection to verify the exact values (initialHeight=700, cornerRadius=16, etc.). The handler tests now just verify the routing — that withCustomTabsOptions gets called or not based on the input.

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
utkrishtsahu: Addressed — same approach as above. Value-level assertions live in CustomTabsOptionsBuilderTest.

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
utkrishtsahu: Fixed in the new utility test. For example, the precedence test asserts that the packages list is exactly ["com.android.chrome"] (from customTabsOptions) and not ["org.mozilla.firefox"] (from top-level). Same for the fallback test — it asserts the exact list passed via top-level allowedBrowsers.

auth0_flutter/android/src/test/kotlin/com/auth0/auth0_flutter/LoginWebAuthRequestHandlerTest.kt
utkrishtsahu: Fixed in the new utility test. For example, the precedence test asserts that the packages list is exactly ["com.android.chrome"] (from customTabsOptions) and not ["org.mozilla.firefox"] (from top-level). Same for the fallback test — it asserts the exact list passed via top-level allowedBrowsers.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/utils/customTabsOptionsBuilder.kt
```diff
@@ -0,0 +1,70 @@
...
+ */
+fun buildCustomTabsOptions(args: Map<*, *>): CustomTabsOptions? {
+    val customTabsOptionsMap = args["customTabsOptions"] as? Map<*, *>
+
+    // Resolve allowedBrowsers: customTabsOptions takes precedence over top-level
+    val allowedBrowsers = if (customTabsOptionsMap != null) {
+        (customTabsOptionsMap["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
+    } else {
+        (args["allowedBrowsers"] as? List<*>)?.filterIsInstance<String>().orEmpty()
+    }
```
pmathew92, line 22: There is still a edge-case where if user passes customTabsOptionsMap !=null but doesn't contain allowed browser list and args["allowedBrowsers"] is passed with a valid list. In this scenario, we would completely skip extracting the allowedBrowsers. Not a hard requirement . Skip if you think this won't happen or chance are very low

pmathew92 (approved):

PR #850, opened 2026-06-04
feat: add Flutter Multi-Resource Refresh Token (MRRT) support

utkrishtsahu:

### 📋 Changes

This PR adds Multi-Resource Refresh Token (MRRT) support to the credentials manager, bringing the Flutter plugin to parity with react-native-auth0, Auth0.Android, and Auth0.swift, which already expose this capability.

**Why:** Previously the Flutter plugin had no way to obtain an access token for an API (audience) other than the one the user logged in with. Apps had to either log in again interactively against the second audience or misuse credentials(parameters: {'audience': ...}), which does not perform a real MRRT exchange (it returns the cached app-audience token when still valid). This adds the dedicated, correct API.

**New public API on CredentialsManager:**

```
Future<ApiCredentials> getApiCredentials({
  required String audience,
  Set<String> scope = const {},
  int minTtl = 0,
  Map<String, String> parameters = const {},
  Map<String, String> headers = const {}, // iOS/macOS only
});

Future<void> clearApiCredentials({
  required String audience,
  String? scope,
});
```

- getApiCredentials exchanges the stored refresh token for an access token scoped to the requested audience (MRRT). If valid API credentials for that audience are already cached, they are returned without a network call; otherwise a new token is fetched and cached per audience.
- clearApiCredentials removes the cached API credentials for an audience (e.g. on logout).

**Layers touched:** platform interface (ApiCredentials model, two options classes, abstract methods, method channel, barrel exports) → Dart public API → Android request handlers → iOS/macOS method handlers → docs (EXAMPLES.md, README.md).

**No breaking changes:** everything is additive (new methods/types; new platform-interface methods follow the existing throw UnimplementedError() default-body pattern). No native dependency bump required — MRRT is available in the already-pinned Auth0.Android 3.16.0 (added in 3.7.0) and Auth0.swift 2.18.0 (added in 2.12.0).

**Prerequisites for use:** Multi-Resource Refresh Tokens must be enabled on the tenant, and the offline_access scope must be requested at login.

### 📎 References

SDK-9521

### 🎯 Testing

**Automated (added in this PR):**

- Dart – platform interface (api_credentials_test.dart, method_channel_credentials_manager_test.dart): ApiCredentials.fromMap/toMap, options toMap (with/without optional args, scope omission), and method-channel routing/result-parsing/error paths for both getApiCredentials and clearApiCredentials.
- Dart – auth0_flutter (credentials_manager_test.dart): both methods pass arguments through to the platform, apply defaults, and return the platform result.
- Android (GetApiCredentialsRequestHandlerTest.kt, ClearApiCredentialsRequestHandlerTest.kt): native call invoked with correct audience/scope/options, success field mapping, failure path, missing-audience error.
- iOS/macOS (CredentialsManagerApiCredentialsMethodHandlerTests.swift, CredentialsManagerClearApiCredentialsMethodHandlerTests.swift): required-argument errors, exchange called with audience/scope, produced dictionary shape, and CredentialsManagerError propagation.

## Summary by CodeRabbit

* **New Features**
  * Added getApiCredentials() and clearApiCredentials() and an ApiCredentials model for per-audience API credential management (MRRT), with options for scopes, TTL, parameters, and headers.

* **Documentation**
  * Updated docs and examples with Credentials Manager MRRT usage, API examples, prerequisites, and platform cache-keying behavior.

* **Tests**
  * Added unit and platform tests covering fetch/clear flows, defaults, error handling, and credentials serialization.

### DIFF
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -726,21 +726,21 @@ final apiCredentials = await auth0.credentialsManager.getApiCredentials(
   scope: {'read:data', 'write:data'},
   minTtl: 60,
   parameters: {'key': 'value'},
-  headers: {'key': 'value'}, // iOS and macOS only
+  headers: {'key': 'value'},
 );
 ```
 
 To remove the cached API credentials for an audience – for example, on logout:
 
 ```dart
-await auth0.credentialsManager.clearApiCredentials(
+final removed = await auth0.credentialsManager.clearApiCredentials(
   audience: 'https://my-api.example.com',
 );
 ```
 
 > ⚠️ **Prerequisites:** Multi-Resource Refresh Tokens must be enabled on your tenant, and the `offline_access` scope must have been requested at login so that a refresh token is available for the exchange.
 >
-> 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them. On Android they are keyed by audience alone, and the `scope` argument is ignored.
+> 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them, and the returned `bool` reflects whether matching credentials were found and removed. On Android they are keyed by audience alone, the `scope` argument is ignored, and the call always returns `true`.
 
 ### Retrieve user profile
 
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -739,7 +739,7 @@ await auth0.credentialsManager.clearApiCredentials(
 ```
 
 > ⚠️ **Prerequisites:** Multi-Resource Refresh Tokens must be enabled on your tenant, and the `offline_access` scope must have been requested at login so that a refresh token is available for the exchange.
-
+>
 > 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them. On Android they are keyed by audience alone, and the `scope` argument is ignored.
 
 ### Retrieve user profile
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -733,14 +733,15 @@ final apiCredentials = await auth0.credentialsManager.getApiCredentials(
 To remove the cached API credentials for an audience – for example, on logout:
 
 ```dart
-final removed = await auth0.credentialsManager.clearApiCredentials(
+await auth0.credentialsManager.clearApiCredentials(
   audience: 'https://my-api.example.com',
+  scope: 'read:data write:data',
 );
 ```
 
 > ⚠️ **Prerequisites:** Multi-Resource Refresh Tokens must be enabled on your tenant, and the `offline_access` scope must have been requested at login so that a refresh token is available for the exchange.
 >
-> 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them, and the returned `bool` reflects whether matching credentials were found and removed. On Android they are keyed by audience alone, the `scope` argument is ignored, and the call always returns `true`.
+> 💡 Stored API credentials are keyed by **both** audience and scope on every platform, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them. The native APIs do not consistently report whether a matching entry existed, so this method returns `void` rather than a success flag.
 
 ### Retrieve user profile
 
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -670,6 +670,7 @@ await auth0.windowsWebAuthentication().logout(
 
 - [Check for stored credentials](#check-for-stored-credentials)
 - [Retrieve stored credentials](#retrieve-stored-credentials)
+- [Retrieve API credentials for a specific audience (MRRT)](#retrieve-api-credentials-for-a-specific-audience-mrrt)
 - [Retrieve user profile](#retrieve-user-profile)
 - [Custom implementations](#custom-implementations)
 - [Local authentication](#local-authentication)
@@ -704,6 +705,42 @@ final credentials = await auth0.credentialsManager.credentials();
 ```
 
 > 💡 You do not need to call `credentialsManager.storeCredentials()` afterward. The Credentials Manager automatically persists the renewed credentials.
+
+### Retrieve API credentials for a specific audience (MRRT)
+
+If your app needs an access token for a **different API** than the one it logged in with, use `getApiCredentials()`. It exchanges the stored refresh token for an access token scoped to the requested `audience` using a [Multi-Resource Refresh Token (MRRT)](https://auth0.com/docs/secure/tokens/refresh-tokens/multi-resource-refresh-token). If valid API credentials for that audience are already cached, they are returned without a network call; otherwise a new token is fetched and cached for next time.
+
+```dart
+final apiCredentials = await auth0.credentialsManager.getApiCredentials(
+  audience: 'https://my-api.example.com',
+);
+
+print('Access token: ${apiCredentials.accessToken}');
+```
+
+You can request specific scopes and pass additional options:
+
+```dart
+final apiCredentials = await auth0.credentialsManager.getApiCredentials(
+  audience: 'https://my-api.example.com',
+  scope: {'read:data', 'write:data'},
+  minTtl: 60,
+  parameters: {'key': 'value'},
+  headers: {'key': 'value'}, // iOS and macOS only
+);
+```
+
+To remove the cached API credentials for an audience – for example, on logout:
+
+```dart
+await auth0.credentialsManager.clearApiCredentials(
+  audience: 'https://my-api.example.com',
+);
+```
+
+> ⚠️ **Prerequisites:** Multi-Resource Refresh Tokens must be enabled on your tenant, and the `offline_access` scope must have been requested at login so that a refresh token is available for the exchange.
+
+> 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them. On Android they are keyed by audience alone, and the `scope` argument is ignored.
 
 ### Retrieve user profile
 
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
@@ -20,7 +20,7 @@ class ClearApiCredentialsRequestHandler : CredentialsManagerRequestHandler {
             return
         }
 
-        val scope = request.data["scope"] as String?
+        val scope = request.data["scope"] as? String
 
         // Stored API credentials are keyed by audience and (optional) scope.
         credentialsManager.clearApiCredentials(audience, scope)
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
@@ -20,9 +20,10 @@ class ClearApiCredentialsRequestHandler : CredentialsManagerRequestHandler {
             return
         }
 
-        // The stored API credentials on Android are keyed by audience alone;
-        // the optional `scope` argument (iOS/macOS only) is intentionally ignored.
-        credentialsManager.clearApiCredentials(audience)
+        val scope = request.data["scope"] as String?
+
+        // Stored API credentials are keyed by audience and (optional) scope.
+        credentialsManager.clearApiCredentials(audience, scope)
         result.success(true)
     }
 }
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
@@ -14,8 +14,8 @@ class ClearApiCredentialsRequestHandler : CredentialsManagerRequestHandler {
         request: MethodCallRequest,
         result: MethodChannel.Result
     ) {
-        val audience = request.data["audience"] as String?
-        if (audience == null) {
+        val audience = request.data["audience"] as? String
+        if (audience.isNullOrBlank()) {
             result.error("UNKNOWN ERROR", "audience is required to clear API credentials", null)
             return
         }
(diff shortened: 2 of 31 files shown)


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
```diff
@@ -0,0 +1,28 @@
...
+        credentialsManager: SecureCredentialsManager,
+        context: Context,
+        request: MethodCallRequest,
+        result: MethodChannel.Result
+    ) {
+        val audience = request.data["audience"] as String?
+        if (audience == null) {
+            result.error("UNKNOWN ERROR", "audience is required to clear API credentials", null)
+            return
+        }
```
pmathew92, line 21: This should also take scope as an optional input. Users can have multilple tokens with the same audience but different scope eg : MyAccount

auth0_flutter/example/ios/Tests/CredentialsManager/CredentialsManagerApiCredentialsMethodHandlerTests.swift
```diff
@@ -0,0 +1,108 @@
...
+    func testProducesErrorWhenRequiredArgumentsAreMissing() {
+        let keys: [Argument] = [.audience, .scopes, .minTtl, .parameters, .headers]
+        let expectations = keys.map { expectation(description: "\($0.rawValue) is missing") }
+        for (argument, currentExpectation) in zip(keys, expectations) {
+            sut.handle(with: arguments(without: argument)) { result in
+                assert(result: result, isError: .requiredArgumentMissing(argument.rawValue))
+                currentExpectation.fulfill()
+            }
+        }
+        wait(for: expectations)
```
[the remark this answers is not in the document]
utkrishtsahu: Intentional — leaving as is. The handler mirrors the existing SSOCredentialsMethodHandler, which also guards every channel argument as required. The Dart layer always sends defaults for the optional fields (scope=[], minTtl=0, parameters={}, headers={}), so at the native boundary these keys are always present. The test correctly locks in that actual cross-layer contract; relaxing it would diverge from the established SSO convention without changing real behavior. The optionality the contract guarantees is enforced at the Dart API surface (getApiCredentials({required audience, ...})), which has its own tests.

auth0_flutter/lib/src/mobile/credentials_manager.dart
```diff
@@ -48,6 +48,41 @@ abstract class CredentialsManager {
...
+  /// Removes the API credentials for the given [audience] from the storage if
+  /// present.
+  ///
+  /// On iOS and macOS, pass the [scope] the credentials were stored with. On
+  /// Android the stored credentials are keyed by audience alone, so [scope] is
+  /// ignored there.
+  Future<void> clearApiCredentials({
+    required final String audience,
+    final String? scope,
+  });
```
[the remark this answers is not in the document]
utkrishtsahu: Leaving as is. CredentialsManager is this package's own abstraction with a single implementation (DefaultCredentialsManager). The recently added ssoCredentials was introduced on this same abstract class as an abstract member with no default body — getApiCredentials/clearApiCredentials follow that exact precedent for consistency. Adding throwing default bodies here would be inconsistent with the rest of the class and isn't warranted given there are no external implementers.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
utkrishtsahu: this is already handled in the current revision. ClearApiCredentialsRequestHandler reads the optional scope and passes it through as clearApiCredentials(audience, scope), and the Dart API exposes clearApiCredentials({required String audience, String? scope}). One platform note (also documented in EXAMPLES.md): on iOS/macOS the stored credentials are keyed by audience and scope, so the scope is used there; on Android the native key is audience-only, so the scope argument is accepted but ignored. This covers the MyAccount case of multiple tokens sharing an audience with different scopes on iOS/macOS.

auth0_flutter_platform_interface/lib/src/credentials-manager/method_channel_credentials_manager.dart
```diff
@@ -114,6 +118,36 @@ class MethodChannelCredentialsManager extends CredentialsManagerPlatform {
...
+    return ApiCredentials.fromMap(result);
+  }
+
+  /// Removes the API credentials for the given audience from the native
+  /// storage if present.
+  ///
+  /// Uses the [MethodChannel] to communicate with the native platforms.
+  /// See `CredentialsManager.clearApiCredentials` for full documentation.
+  @override
+  Future<void> clearApiCredentials(
```
sanchitmehtagit, line 142: clearApiCredentials:                                                                                                            
  - Swift: callback(self.credentialsManager.clear(forAudience: audience, scope: scope)) - also passes the Bool back                        
  - Dart return type: Future<void> - discards it                                                                                           
  - Caller has no way to know if the audience was found and cleared 

I think clearCredentials also return bool

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
```diff
@@ -0,0 +1,29 @@
...
+        request: MethodCallRequest,
+        result: MethodChannel.Result
+    ) {
+        val audience = request.data["audience"] as? String
+        if (audience.isNullOrBlank()) {
+            result.error("UNKNOWN ERROR", "audience is required to clear API credentials", null)
+            return
+        }
+
+        val scope = request.data["scope"] as String?
```
pmathew92:
```suggestion
        val scope = request.data["scope"] as? String
```

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+class GetApiCredentialsRequestHandler : CredentialsManagerRequestHandler {
+    override val method: String = "credentialsManager#getApiCredentials"
+
+    override fun handle(
+        credentialsManager: SecureCredentialsManager,
+        context: Context,
+        request: MethodCallRequest,
+        result: MethodChannel.Result
+    ) {
+        val audience = request.data["audience"] as String?
```
pmathew92:
```suggestion
        val audience = request.data["audience"] as? String
```

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+            return
+        }
+
+        var scope: String? = null
+        val scopes = (request.data["scopes"] ?: arrayListOf<String>()) as ArrayList<*>
+        if (scopes.isNotEmpty()) {
+            scope = scopes.joinToString(separator = " ")
+        }
+
+        val minTtl = request.data["minTtl"] as Int? ?: 0
```
pmathew92:
```suggestion
        val minTtl = request.data["minTtl"] as? Int ?: 0
```

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+        val parameters = request.data["parameters"] as Map<String, String>? ?: mapOf()
```
pmathew92:
```suggestion
        val parameters = request.data["parameters"] as? Map<String, String> ?: mapOf()
```

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+        val headers = request.data["headers"] as Map<String, String>? ?: mapOf()
```
pmathew92:
```suggestion
        val headers = request.data["headers"] as? Map<String, String> ?: mapOf()
```

auth0_flutter/EXAMPLES.md
```diff
@@ -705,6 +706,42 @@ final credentials = await auth0.credentialsManager.credentials();
...
+
+You can request specific scopes and pass additional options:
+
+```dart
+final apiCredentials = await auth0.credentialsManager.getApiCredentials(
+  audience: 'https://my-api.example.com',
+  scope: {'read:data', 'write:data'},
+  minTtl: 60,
+  parameters: {'key': 'value'},
+  headers: {'key': 'value'}, // iOS and macOS only
```
pmathew92: ```headers: {'key': 'value'}``` is valid on Android also. remove the iOS and macOS only line

auth0_flutter_platform_interface/lib/src/credentials-manager/method_channel_credentials_manager.dart
utkrishtsahu: Good catch — fixed. clearApiCredentials now returns Future<bool> end-to-end (method channel → platform interface → public API), mirroring clearCredentials, so the caller can tell whether matching credentials were found and removed. One platform note: on iOS/macOS this is the real Bool from clear(forAudience:scope:); on Android the native clearApiCredentials is void, so it returns true best-effort (documented in EXAMPLES.md). Added return-value assertions to the Dart tests.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/ClearApiCredentialsRequestHandler.kt
utkrishtsahu: Done — changed to as? String.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
utkrishtsahu: Done — changed to as? String.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
utkrishtsahu: Done — changed to as? Int ?: 0.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
utkrishtsahu: Done — changed to as? Map<String, String> ?: mapOf().

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/credentials_manager/GetApiCredentialsRequestHandler.kt
utkrishtsahu: Done — changed to as? Map<String, String> ?: mapOf().

auth0_flutter/EXAMPLES.md
utkrishtsahu: Good catch — removed the comment. Auth0.Android's getApiCredentials accepts headers too, so the note was inaccurate.

sanchitmehtagit (dismissed):

auth0_flutter/EXAMPLES.md
```diff
@@ -705,6 +706,42 @@ final credentials = await auth0.credentialsManager.credentials();
...
+```dart
+final removed = await auth0.credentialsManager.clearApiCredentials(
+  audience: 'https://my-api.example.com',
+);
+```
+
+> ⚠️ **Prerequisites:** Multi-Resource Refresh Tokens must be enabled on your tenant, and the `offline_access` scope must have been requested at login so that a refresh token is available for the exchange.
+>
+> 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them, and the returned `bool` reflects whether matching credentials were found and removed. On Android they are keyed by audience alone, the `scope` argument is ignored, and the call always returns `true`.
+
```
sanchitmehtagit, line 745: Just wanted to confirm: does this API always return true on Android. Is that expected?

pmathew92: No that is not expected. Android APIs don't return any confirmation for clear scenario

auth0_flutter/EXAMPLES.md
```diff
@@ -705,6 +706,42 @@ final credentials = await auth0.credentialsManager.credentials();
...
+```dart
+final removed = await auth0.credentialsManager.clearApiCredentials(
+  audience: 'https://my-api.example.com',
+);
+```
+
+> ⚠️ **Prerequisites:** Multi-Resource Refresh Tokens must be enabled on your tenant, and the `offline_access` scope must have been requested at login so that a refresh token is available for the exchange.
+>
+> 💡 On iOS and macOS, the stored API credentials are keyed by audience **and** scope, so pass the same `scope` to `clearApiCredentials()` that you used when fetching them, and the returned `bool` reflects whether matching credentials were found and removed. On Android they are keyed by audience alone, the `scope` argument is ignored, and the call always returns `true`.
```
pmathew92: Two things :
1 : ` On Android they are keyed by audience alone, the `scope` argument is ignored`
Wrong. Both Android and iOS scope it to audience and scope.  So user need to pass both values to clear the correct token else they will delete all the tokens for that audience . Please correct the above
2: Android APIs return void for clearCredentials API incase of both failure and success.  Returning `true` always conveys a false positive even when the failure failed for some reason . Lets not add the ` always returns `true` ` text

pmathew92 (changes requested):

auth0_flutter/lib/src/mobile/credentials_manager.dart
```diff
@@ -48,6 +48,43 @@ abstract class CredentialsManager {
...
+
+  /// Removes the API credentials for the given [audience] from the storage if
+  /// present.
+  ///
+  /// On iOS and macOS, pass the [scope] the credentials were stored with. On
+  /// Android the stored credentials are keyed by audience alone, so [scope] is
+  /// ignored there.
+  ///
+  /// Returns `true` if credentials for the [audience] were found and removed.
+  Future<bool> clearApiCredentials({
```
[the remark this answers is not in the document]
pmathew92: Valid suggestion .Android API doesn't return true but void . Returning true would be a wrong guarantee here

pmathew92 (dismissed):

sanchitmehtagit (approved):

PR #864, opened 2026-06-11
feat: Add Multi-Resource Refresh Token (MRRT) support for Flutter Web

utkrishtsahu:

### 📋 Changes

This PR adds Multi-Resource Refresh Token (MRRT) support to the Flutter Web platform, bringing it to parity with the native mobile support added built on @auth0/auth0-spa-js (MRRT landed in spa-js v2.5.0; the example already pins 2.9.0).

MRRT lets a single refresh token be exchanged for per-audience access tokens without re-authentication.

**Public API additions (Auth0Web):**

- onLoad({..., bool? useMrrt}) — new client option. Enables MRRT on the underlying spa-js client and auto-enables refresh tokens when useMrrt is true and useRefreshTokens is not set explicitly (useRefreshTokens ?? useMrrt), since MRRT requires refresh tokens.
await auth0Web.onLoad(useMrrt: true);
- Future<ApiCredentials> getApiCredentials({required String audience, Set<String> scopes, Map<String, String> parameters}) — exchanges the refresh token for an access token scoped to audience. Wraps getTokenSilently({authorizationParams: {audience, scope, ...parameters}, detailedResponse: true}) and maps the result to the existing ApiCredentials model.
final creds = await auth0Web.getApiCredentials(
  audience: 'https://my.api',
  scopes: {'read:messages'},
);
- Future<void> clearApiCredentials({required String audience, String? scope}) — documented no-op kept for cross-platform API parity. auth0-spa-js manages its own per-audience cache and exposes no single-audience evict API; this logs a console.warn and resolves.

**Internal changes:**

- useMrrt plumbed through all four web layers (JS interop → ClientOptions model → client-options extension → public onLoad), mirroring the existing useDPoP flag.
- New ApiCredentialsExtension.fromWeb maps spa-js WebCredentials → ApiCredentials (computes expiresAt from expires_in, splits scope).
- New abstract methods on Auth0FlutterWebPlatform (default UnimplementedError), implemented in the real web plugin.

### 📎 References

SDK-9561

### 🎯 Testing

New/updated coverage:

- client_options_extension_test.dart — useMrrt: true is forwarded to spa-js and auto-enables useRefreshTokens; an explicit useRefreshTokens: false is preserved; useMrrt is omitted by default.
- api_credentials_extension_test.dart — fromWeb mapping with required values (defaults tokenType to Bearer, computes expiresAt) and optional values (scope, token_type).
- auth0_flutter_web_test.dart —
- getApiCredentials calls getTokenSilently with the correct authorizationParams (audience + space-joined scope) and detailedResponse: true, and maps the response to ApiCredentials.
- getApiCredentials wraps a JS error into WebException.
- clearApiCredentials resolves without throwing and does not call the spa-js client.

## Summary by CodeRabbit

* **New Features**
  * Added web APIs to fetch per-audience API credentials with optional scopes and parameters, and to convert results into `ApiCredentials`.
  * Added web API to clear API credentials (web is a no-op aside from signaling; storage is handled automatically).
  * Introduced MRRT support via new `useMrrt` option, including startup behavior and forcing refresh-token usage when enabled.
  * Re-exported `ApiCredentials` on the web surface.
* **Tests**
  * Added web tests for credential retrieval/clearing, `ApiCredentials` conversion, and MRRT/client option forwarding behavior.


auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -173,6 +174,40 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
...
+    }
+  }
+
+  @override
+  Future<void> clearApiCredentials(
+      final ClearApiCredentialsOptions options) async {
+    final scopeInfo =
+        options.scope != null ? ' and scope ${options.scope}' : '';
+    final message =
+        "'clearApiCredentials' for audience ${options.audience}$scopeInfo "
```
NandanPrabhu: do we want to return with clearApiCredentials for audience and scope?

auth0_flutter/lib/src/web/extensions/client_options_extensions.dart
```diff
@@ -19,9 +19,10 @@ extension ClientOptionsExtension on ClientOptions {
           sessionCheckExpiryDays: sessionCheckExpiryInDays,
           useCookiesForTransactions: useCookiesForTransactions,
           useFormData: useFormData,
-          useRefreshTokens: useRefreshTokens,
+          useRefreshTokens: useRefreshTokens ?? useMrrt,
```
pmathew92: What if user passes false for `useRefreshTokens` and true for `useMrrt` ?

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
pmathew92: Agree you don't need to log the audience and scope here. Just message saying this operation is not supported on web since auth0-spa-js handles credential storage automatically.

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -66,6 +66,12 @@ class Auth0Web {
   /// to learn more.
   /// * [scopes] defaults to `openid profile email`. You can override these
   /// scopes, but `openid` is always requested regardless of this setting.
+  /// * [useMrrt] enables Multi-Resource Refresh Tokens, allowing a single
+  /// refresh token to be reused to obtain access tokens for multiple APIs
+  /// (audiences). Once enabled, request per-audience tokens via
+  /// [getApiCredentials] or [credentials]. Enabling this implicitly enables
```
pmathew92:
```suggestion
  /// [getApiCredentials]. Enabling this implicitly enables
```

auth0_flutter/lib/auth0_flutter_web.dart
```diff
@@ -369,6 +377,48 @@ class Auth0Web {
...
+  }) =>
+      Auth0FlutterWebPlatform.instance.getApiCredentials(
+        GetApiCredentialsOptions(
+          audience: audience,
+          scopes: scopes,
+          parameters: parameters,
+        ),
+      );
+
+  /// Removes the stored API credentials for the given [audience].
```
pmathew92: We need to document this better here since this operation is not supported on Web.

utkrishtsahu:
Note: on the failing Run native Windows unit tests job — unrelated to this PR.

This PR is entirely Dart/web code (auth0_flutter/lib/src/web/** + platform-interface). All Dart/Flutter CI jobs pass — Analyze auth0_flutter, Test auth0_flutter, and Test auth0_flutter_platform_interface are all green. The only red job is the Windows native C++ build, which never touches web code.

The failure is pre-existing toolchain drift, not a regression here:

The same commit on this branch was green on Jun 11 (run 27325784210) and red on Jun 15 (run 27537023026) — no code changed in between.
windows-latest rolled its compiler forward to VS 18 / MSVC 19.51, which removed stdext::checked_array_iterator. The pinned (and now EOL) cpprestsdk 2.10.19 uses it, so it hard-errors with C2065: 'stdext' undeclared.
This breaks main and every open PR equally — it's a CI infra issue, not specific to this change.
Tracking the fix separately (pin the Windows runner image; longer-term migrate off the archived cpprestsdk) so it doesn't bloat this web-only PR.

auth0_flutter/lib/src/web/extensions/client_options_extensions.dart
```diff
@@ -19,9 +19,10 @@ extension ClientOptionsExtension on ClientOptions {
           sessionCheckExpiryDays: sessionCheckExpiryInDays,
           useCookiesForTransactions: useCookiesForTransactions,
           useFormData: useFormData,
-          useRefreshTokens: useRefreshTokens,
+          useRefreshTokens: useMrrt == true ? true : useRefreshTokens,
```
pmathew92, line 26: We can simplify it as `useRefreshTokens = useMrrt ? true : useRefreshTokens`

auth0_flutter/lib/src/web/auth0_flutter_plugin_real.dart
```diff
@@ -173,6 +174,37 @@ class Auth0FlutterPlugin extends Auth0FlutterWebPlatform {
...
+      return ApiCredentialsExtension.fromWeb(result);
+    } catch (e) {
+      throw WebExceptionExtension.fromJsObject(JSObject.fromInteropObject(e));
+    }
+  }
+
+  @override
+  Future<void> clearApiCredentials(
+      final ClearApiCredentialsOptions options) async {
+    console.warn(
```
pmathew92, line 202: Should we do a log or throw an exception in this scenario ?

utkrishtsahu: It's an intentional no-op with a warning rather than a throw. This method only exists on web for cross-platform parity — shared code calling clearApiCredentials(audience:) should work on both mobile and web. On web, auth0-spa-js manages its own cache and doesn't expose a per-audience evict API, so there's nothing for us to clear. Throwing would break that shared call on web for no real benefit.

auth0_flutter/lib/src/web/extensions/client_options_extensions.dart
utkrishtsahu: looked at this — useMrrt is bool?, so useMrrt ? true : useRefreshTokens won't compile (Dart rejects a nullable bool as a condition). The == true is the null-guard.

I also considered useRefreshTokens ?? useMrrt ?? false (the rn-auth0 form), but that changes the behavior: if someone passes useMrrt: true with useRefreshTokens: false, it'd return false and silently break MRRT. The current form deliberately forces refresh tokens on whenever MRRT is enabled, which is what the useMrrt doc comment guarantees. Added a comment to make that intent clear.

pmathew92: simple if else will give better reading clarity , 
`useRefreshTokens: if(useMrrt) useMrrt else useRefreshTokens`
Your call

pmathew92 (approved):

PR #872, opened 2026-06-19
feat: Add Multi-Factor Authentication (MFA) support for mobile

utkrishtsahu:

### 📋 Changes
This PR adds **Multi-Factor Authentication (MFA) support** to the Flutter SDK for **iOS and Android**, built on the flexible-grant MFA APIs in Auth0.swift and Auth0.Android. It lets apps complete MFA flows after receiving an `mfa_required` error during authentication — enrolling new factors, challenging existing ones, and exchanging a verification code for credentials.

**New public API — `auth0.mfa` (`MfaApi`):**

| Method | Description |
|---|---|
| `getAuthenticators(...)` | Lists the user's enrolled authenticators |
| `enrollTotp(...)` | Enrolls a TOTP (authenticator app) factor; returns secret + barcode URI for QR rendering |
| `enrollPhone(...)` | Enrolls an SMS/Voice (OOB) factor for an E.164 phone number |
| `enrollEmail(...)` | Enrolls an Email (OOB) factor |
| `enrollPush(...)` | Enrolls a Push (Guardian) factor; returns a barcode URI to scan |
| `challenge(...)` | Requests a challenge against an enrolled authenticator |
| `verify(...)` | Verifies an `otp`, `oob`, or `recoveryCode` grant and returns `Credentials` |

All methods take the `mfaToken` from a `mfa_required` authentication error.

**Platform interface (`auth0_flutter_platform_interface`):**
- New MFA method channel (`auth0.com/auth0_flutter/mfa`) and platform contract.
- New models: `MfaAuthenticator`, `MfaChallenge`, `MfaEnrollmentChallenge`, `MfaRequirements`, the `Mfa*Options` request types, and `MfaException`.
- `mfa_required` challenge details are now surfaced on `ApiException`.

**Native implementations:**
- **iOS/macOS:** new `MfaAPI/` method handlers built on the `MFAClient` API. Bumps Auth0.swift to `2.23.0` (all three podspecs).
- **Android:** new `request_handlers/mfa/` request handlers on `MfaApiClient`. Bumps `com.auth0.android:auth0` to `3.21.0`.

**Behavioral change — `getAuthenticators(factorsAllowed:)` now filters on `Authenticator.type` exactly:**
- `factorsAllowed` is matched **exactly** against each authenticator's `type` (e.g. `totp` / `phone` / `email` / `recovery-code`), **not** the higher-level `otp` / `oob` categories.
- Aligns both platforms with Auth0.Android 3.21.0 ([#998](https://github.com/auth0/Auth0.Android/pull/998)) and removes the previous iOS `expandFactors` workaround, so `factorsAllowed` is now passed through unchanged on both platforms.
- Callers who previously passed `otp` / `oob` must switch to the resolved `.type` values (the same values exposed as `MfaFactor.type` on `mfaRequirements`).

**Other fixes:**
- **Android TOTP enrollment:** reads the manual-entry key from `secret` (per 3.21.0's `TotpEnrollmentChallenge`, where `manualInputCode` is `null` on the `/mfa/associate` path) and surfaces `recovery_codes`.
- **Web:** `mfa_required` error parsing now reads the token from `mfa_token` (auth0-spa-js exposes it snake_cased), fixing a crash where the token surfaced as `type 'Null' is not a subtype of type 'String'`.

**Docs:** MFA usage examples added to `EXAMPLES.md`, updated to the `.type` `factorsAllowed` vocabulary.

**Out of scope:** Desktop platforms and Push auto-polling are intentionally not included.

### 📎 References

SDK-6406

### 🎯 Testing

**Automated tests included:**
- **Dart:** unit tests for `MfaApi` (`mfa_api_test.dart`), the method channel (`method_channel_auth0_flutter_mfa_test.dart`), and `mfa_required` parsing on `ApiException` (`api_exception_test.dart`).
- **Android:** Kotlin unit tests for every request handler and the MFA method-call handler (`request_handlers/mfa/*Test.kt`), including the TOTP `secret` / `recovery_codes` mapping.
- **iOS:** Swift unit tests for every method handler with spies (`example/ios/Tests/Mfa/*`), including `getAuthenticators` pass-through of `factorsAllowed`.

**Manual end-to-end verification:**
- **Android** (Pixel 7a) and **iOS** (iPhone 16 Pro sim): exact `.type` filtering confirmed (no `oob` widening), TOTP enroll returns a non-null secret matching the QR, full challenge → verify round-trip.
- **Web** (spa-js flexible-grant): `mfa_required` → `getAuthenticators` (server-side `.type` filtering) → challenge → `verifyOob` round-trip completed successfully.


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MfaExtensions.kt
```diff
@@ -0,0 +1,72 @@
...
+        "idToken" to idToken,
+        "refreshToken" to refreshToken,
+        "userProfile" to user.toMap(),
+        "expiresAt" to formattedDate,
+        "scopes" to scopes,
+        "tokenType" to type
+    )
+}
+
+fun MfaException.toMfaMap(): Map<String, Any> = buildMap {
```
pmathew92, line 66: All type of MFAExceptions has code  and description too which is what actually carries the error . Add those two

auth0_flutter/lib/src/mobile/mfa_api.dart
```diff
@@ -0,0 +1,222 @@
...
+  ///
+  /// ## Usage example
+  ///
+  /// ```dart
+  /// final credentials = await mfa.verifyOob(
+  ///   oobCode: challenge.oobCode!,
+  ///   bindingCode: '123456',
+  /// );
+  /// ```
+  Future<Credentials> verifyOob({
```
pmathew92, line 222: We need to have an option to pass scope and audience for methods returning Credentials. The bridge layer should pass the scope and audience to the corresponding native and SPA SDKs

auth0_flutter/lib/src/mobile/mfa_api.dart
```diff
@@ -0,0 +1,222 @@
...
+  /// [Credentials].
+  ///
+  /// ## Usage example
+  ///
+  /// ```dart
+  /// final credentials = await mfa.verifyRecoveryCode(
+  ///   recoveryCode: 'ABCD1234...',
+  /// );
+  /// ```
+  Future<Credentials> verifyRecoveryCode(
```
pmathew92: Same as https://github.com/auth0/auth0-flutter/pull/872/changes#r3466104274

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/MfaExtensions.kt
utkrishtsahu: done. MfaException.toMfaMap() now surfaces both code and description (the actual error carried by the native SDK) in the details map, in addition to the top-level result.error code/message:


put("code", getCode())
put("description", getDescription())
I also mirrored this on iOS/macOS (MfaExtensions.swift) so the Dart layer gets the same code/description contract on every platform.

auth0_flutter/lib/src/mobile/mfa_api.dart
utkrishtsahu: added optional scopes and audience params to the three verify methods that return Credentials (verifyOtp, verifyOob, verifyRecoveryCode). They flow through MfaVerifyOptions and the bridge passes them down to each platform:

Android — appended to the verify token request via Request.addParameter("scope", …) / addParameter("audience", …)
iOS/macOS — chained onto the request with Request.parameters(...)
Web — passed to auth0-spa-js as scope (space-delimited) and audience on MfaVerifyParams
scopes is a Set<String> joined with spaces to match the existing token-exchange convention (audience is only sent when non-null). Added unit tests on all four layers (platform-interface serialization, mobile, web, Android handler, plus an iOS smoke test). The methods that don't return Credentials (getAuthenticators, challenge, enroll*) were left untouched.

auth0_flutter/lib/src/mobile/mfa_api.dart
utkrishtsahu: MfaException.toMfaMap() now includes the code and description from the exception (they carry the actual error), alongside the existing status code and error flags. This was pushed in an earlier commit.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/mfa/VerifyRequestHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+            "recovery_code" -> {
+                assertHasProperties(listOf("recoveryCode"), request.data)
+                MfaVerificationType.RecoveryCode(request.data["recoveryCode"] as String)
+            }
+            else -> throw IllegalArgumentException("Unknown grantType: $grantType")
+        }
+
+        val verifyRequest = client.verify(type)
+
+        val scopes = (request.data["scopes"] ?: arrayListOf<String>()) as ArrayList<*>
```
pmathew92, line 50: Shouldn't this be default scope values of "openid email profile"

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterMfaMethodCallHandler.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.auth0.auth0_flutter.utils.assertHasProperties
+import io.flutter.plugin.common.MethodCall
+import io.flutter.plugin.common.MethodChannel.MethodCallHandler
+import io.flutter.plugin.common.MethodChannel.Result
+
+class Auth0FlutterMfaMethodCallHandler(
+    private val mfaRequestHandlers: List<MfaRequestHandler>
+) : MethodCallHandler {
+
+    override fun onMethodCall(@NonNull call: MethodCall, @NonNull result: Result) {
```
pmathew92, line 16: currently dpop handling is not added for MFA. Create a ticket to track this

auth0_flutter/lib/src/mobile/mfa_api.dart
```diff
@@ -0,0 +1,254 @@
...
+  /// ```dart
+  /// final credentials = await mfa.verifyOtp(
+  ///   otp: '123456',
+  ///   audience: 'https://my-api.example.com',
+  ///   scopes: {'openid', 'profile', 'read:messages'},
+  /// );
+  /// ```
+  Future<Credentials> verifyOtp({
+    required final String otp,
+    final Set<String> scopes = const {},
```
pmathew92: Lets add default scope at the dart level

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/mfa/VerifyRequestHandler.kt
utkrishtsahu: Done. Set the default scopes to {'openid', 'profile', 'email'} at the Dart level on all three verify methods (verifyOtp, verifyOob, verifyRecoveryCode) in both MfaApi (mobile) and MfaWeb (web). Since the default lives in the public Dart signature, it flows through MfaVerifyOptions down to every platform automatically. Updated the dartdoc and tests accordingly.

auth0_flutter/lib/src/mobile/mfa_api.dart
utkrishtsahu: Handled this at the Dart layer (see the comment on mfa_api.dart) so the default is consistent across Android/iOS/web. The native handler keeps its if (scopes.isNotEmpty()) guard — it now always receives openid profile email by default, so no separate Kotlin-side default is needed.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterMfaMethodCallHandler.kt
utkrishtsahu: DPoP isn't applied to the MFA token-exchange path yet. Raised a tracking ticket to add it across all platforms as a follow-up: [SDK-9878]. Keeping it out of this PR to avoid scope creep.

pmathew92 (dismissed):

auth0_flutter/lib/src/mobile/mfa_api.dart
```diff
@@ -0,0 +1,262 @@
...
+  ///
+  /// ```dart
+  /// final authenticators =
+  ///     await mfa.getAuthenticators(factorsAllowed: ['totp', 'phone']);
+  /// ```
+  Future<List<MfaAuthenticator>> getAuthenticators(
+          {required final List<String> factorsAllowed}) =>
+      Auth0FlutterMfaPlatform.instance.getAuthenticators(_createApiRequest(
+          MfaGetAuthenticatorsOptions(
+              mfaToken: _mfaToken, factorsAllowed: factorsAllowed)));
```
pmathew92: The validation comment from @coderabbitai is a valid one. We need to have a check if the underlying native SDKs throw exception for empty list

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/mfa/VerifyRequestHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+        val type = when (val grantType = request.data["grantType"] as String) {
+            "otp" -> {
+                assertHasProperties(listOf("otp"), request.data)
+                MfaVerificationType.Otp(request.data["otp"] as String)
+            }
+            "oob" -> {
+                assertHasProperties(listOf("oobCode"), request.data)
+                MfaVerificationType.Oob(
+                    request.data["oobCode"] as String,
+                    request.data["bindingCode"] as? String
```
pmathew92, line 37: Do safe casting for these properties

auth0_flutter/lib/src/web/extensions/mfa_extensions.dart
```diff
@@ -0,0 +1,59 @@
...
+
+import '../js_interop.dart' as interop;
+
+/// Maps the `factorType` discriminant expected by auth0-spa-js' `mfa.enroll`
+/// for phone factors.
+extension PhoneTypeWebExtension on PhoneType {
+  String toWebFactorType() {
+    switch (this) {
+      case PhoneType.voice:
+        return 'voice';
```
pmathew92, line 13: Does web support `voice` type ?

auth0_flutter/lib/src/mobile/mfa_api.dart
utkrishtsahu: Good catch — added a client-side ArgumentError guard in getAuthenticators (empty list now fails fast instead of a generic server invalid_request), updated docs, and added a test. Web takes no factorsAllowed (spa-js filters server-side), so nothing to change there.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/mfa/VerifyRequestHandler.kt
utkrishtsahu: Done — switched grantType/otp/oobCode/recoveryCode to as? String ?: throw (repo's existing pattern), kept bindingCode optional, dropped the redundant assertHasProperties, and added wrong-type tests.

auth0_flutter/lib/src/web/extensions/mfa_extensions.dart
utkrishtsahu: Yes — spa-js supports it: MfaFactorType = 'otp' | 'sms' | 'email' | 'push' | 'voice'. So 'voice' is correct. Mobile stays SMS-only since the native SDKs always request oob_channels: ['sms'].

pmathew92 (approved):

PR #891, opened 2026-07-09
feat: Add custom token exchange delegation and impersonation support

utkrishtsahu:

### 📋 Changes

This PR adds **Custom Token Exchange (CTE) Delegation & Impersonation** support (RFC 8693), achieving parity with the native SDKs (auth0-spa-js #1608, Auth0.swift #1190, Auth0.Android #975). The base CTE (`customTokenExchange` with a subject token) already shipped; this adds the **actor token** request parameter and exposes the **`act` claim** on the user profile.

**Public API additions:**

- `customTokenExchange` (both `Auth0().api.customTokenExchange` on mobile and `Auth0Web.customTokenExchange` on web) now accepts an optional **`actor`** parameter of type **`ActorToken`** — a value object bundling `token` and `tokenType` (both `required`), so the token and its type can never be partially supplied (pairing is guaranteed by the type at compile time; no runtime validation needed).
- New **`ActorToken`** model (`token`, `tokenType`) representing the acting party in a delegation/impersonation flow.
- New **`UserActor`** model (`sub`, nested `actor`, and `extraClaims`) representing the `act` claim, including nested delegation chains.
- `UserProfile` gains an `actor` field, populated from the `act` claim.
- **Behavioral note:** when an actor token is present, Auth0 does **not** issue a refresh token regardless of the `offline_access` scope, so `credentials.refreshToken` is `null` in this flow. The SDK handles this gracefully.

**Implementation by layer:**

- **Platform interface:** new `ActorToken` and `UserActor` models + exports; `UserProfile.actor`; an optional `actor` (`ActorToken?`) field on the mobile and web option types.
- **Mobile Dart API:** adds the `actor` param on `customTokenExchange`.
- **Android:** handler reconstructs `ActorToken` and calls the actor-aware overload; `UserProfileExtensions` maps the `act` claim recursively (incl. nested chains).
- **iOS/macOS:** handler builds `ActorToken` and calls the `actorToken:` overload; `Extensions`/`Properties` map the `act` claim into `UserInfo`.
- **Web:** `actor_token`/`actor_token_type` are passed through the auth0-spa-js JS interop; the `act` claim is decoded from the returned ID token into `UserProfile.actor`.
- **Docs:** new Delegation & Impersonation section in `EXAMPLES.md`, plus a documented web SDK version requirement in `README.md`.

**SDK version requirements:**

- **iOS/Android:** no native SDK bump required — Auth0.Android `3.20.0` and Auth0.swift `2.23.0` already ship the actor APIs.
- **Web:** actor-token support in `auth0-spa-js` was added in **v2.20.0** (PR #1608). Earlier versions accept the option but silently drop the actor token before the `/oauth/token` request, so the returned ID token has no `act` claim. The README and EXAMPLES now document the **≥ 2.20.0** requirement.

### 📎 References

SDK-8925

### 🎯 Testing

**Automated (added in this PR):**

- **Platform interface:** `ActorToken` and `UserActor` (fromMap/toMap, nested chains, missing `sub`, `extraClaims` precedence), `UserProfile` `act`-claim parsing/emission, and options serialization (wire keys emitted atomically).
- **Mobile Dart:** `customTokenExchange` passes the `actor` token/type through, and leaves them null when no actor is supplied.
- **Web:** actor token passthrough, and `act`-claim decoding into `UserProfile.actor` (excluded from the `customClaims` catch-all).
- **Android:** handler passes `ActorToken` to the SDK (and omits it when absent); safe-casts required args; `UserProfileExtensions` maps the `act` claim incl. nested chains.
- **iOS/macOS:** handler test asserts actor params reach the SDK via the exchange call, plus success/error/optional-argument coverage.

The Dart and Web suites pass locally (`flutter analyze` clean; platform-interface, mobile, and web tests green). The Android and iOS/macOS native tests run in CI.

**Manual end-to-end (verified on a configured tenant, Android + iOS + Web):**

1. On a tenant with custom token exchange + actor support enabled, set up an Action that calls `api.authentication.setActor()`.
2. Call `customTokenExchange` with a valid subject token plus an `actor: ActorToken(...)`.
3. Confirm credentials are returned, that `credentials.user.actor` reflects the expected `act` claim (including nested actors for delegation chains), and that `refreshToken` is `null`.

Verified end-to-end on **Android (Pixel 7a), iOS (iPhone 16 Pro simulator), and Web (Chrome)** — all three return the expected `act` claim (`actor.sub` + `extraClaims`) with refresh-token suppression.

## Summary by CodeRabbit

* **New Features**
  * Added delegation and impersonation support to Custom Token Exchange using actor tokens on Mobile and Web.
  * Exposed actor information through `credentials.user.actor`, including nested delegation chains and additional claims.
  * Actor-token exchanges no longer issue refresh tokens.
* **Documentation**
  * Added usage examples and Web setup guidance, including the required `auth0-spa-js` version for actor claims.


auth0_flutter/lib/src/mobile/authentication_api.dart
```diff
@@ -411,14 +430,24 @@ class AuthenticationApi {
...
-  }) =>
-      Auth0FlutterAuthPlatform.instance.customTokenExchange(_createApiRequest(
-          AuthCustomTokenExchangeOptions(
-              subjectToken: subjectToken,
-              subjectTokenType: subjectTokenType,
-              audience: audience,
-              scopes: scopes,
-              organization: organization)));
+    final String? actorToken,
+    final String? actorTokenType,
```
pmathew92: Can these be taken as class property similar to in native SDKs. In that way you can ensure that both actorToken and actorTokenType are always provided if passed

auth0_flutter/EXAMPLES.md
```diff
@@ -1413,6 +1414,52 @@ final credentials = await auth0Web.customTokenExchange(
...
+
+```dart
+final credentials = await auth0Web.customTokenExchange(
+  subjectToken: 'external-idp-token',
+  subjectTokenType: 'urn:acme:legacy-token',
+  actorToken: 'actor-id-token',
+  actorTokenType: 'urn:ietf:params:oauth:token-type:id_token',
+);
+
+final actor = credentials.user.actor;
```
kishore7snehil, line 1470: This web example reads `credentials.user.actor`, but on web that's always null. On mobile the `act` claim gets mapped onto the actor, but the web path builds the user profile through `UserProfileExtension.fromClaims`, and that only maps the fixed public claims and sweeps every other claim into `customClaims`. It never looks at `act`, so two things happen on web: `.actor` stays null, and the raw `act` claim ends up inside `customClaims` instead.

The requirement is that the `act` claim is reachable through the user object on the exchange response, so web and mobile should behave the same here. Shall we map `act` in `fromClaims` (same shape `UserActor.fromMap` already parses) and drop it from the `customClaims` catch-all, so `.actor` is populated on web and the claim isn't also duplicated into `customClaims`?

auth0_flutter/test/web/auth0_flutter_web_test.dart
```diff
@@ -502,6 +502,44 @@ void main() {
...
+      await auth0.customTokenExchange(
+          subjectToken: 'external-token-456',
+          subjectTokenType: 'urn:example:custom-token',
+          actorToken: 'actor-token-789',
+          actorTokenType: 'urn:ietf:params:oauth:token-type:id_token');
+
+      final options =
+          verify(mockClientProxy.exchangeToken(captureAny)).captured.first;
+      expect(options.actor_token, 'actor-token-789');
+      expect(options.actor_token_type,
```
kishore7snehil, line 520: These web tests cover the request side well (the actor token and type reach the JS interop), but nothing covers the response side, so there's no test that decodes a returned ID token carrying an `act` claim and checks what comes back on the user. That's the gap that let the missing web mapping through.

Can we add a case that returns credentials with an `act` claim in the ID token and asserts `credentials.user.actor` is populated (including a nested actor), and also that `act` isn't left sitting in `customClaims`? That matches what the mobile side already verifies.

kishore7snehil (changes requested):

auth0_flutter/lib/src/mobile/authentication_api.dart
utkrishtsahu: Done. Replaced the two flat actorToken/actorTokenType params with a single typed actor: ActorToken({required token, required tokenType}), matching the native SDKs. Both fields are required on the object, so the token and its type can never be partially supplied — the pairing is now guaranteed at compile time and the previous XOR validation is gone.

auth0_flutter/EXAMPLES.md
utkrishtsahu: Fixed in UserProfileExtension.fromClaims: it now parses act via UserActor.fromMap into .actor and removes act from the customClaims catch-all (added PublicClaims.act to the removed keys). Web and mobile now behave the same — .actor is populated and the claim is no longer duplicated into customClaims.

auth0_flutter/test/web/auth0_flutter_web_test.dart
utkrishtsahu: Added. New test customTokenExchange populates user.actor from the act claim and keeps it out of customClaims decodes a returned ID token carrying a nested act claim and asserts user.actor.sub (actor-agent-123), the nested user.actor.actor.sub (delegated-agent-456) + its claims, and that act isn't left in customClaims — matching the mobile-side coverage.

auth0_flutter/EXAMPLES.md
```diff
@@ -1419,6 +1414,58 @@ final credentials = await auth0Web.customTokenExchange(
...
+
+When present, the resulting ID token may contain an `act` claim (set server-side via an Auth0 Action using `api.authentication.setActor()`), exposed on the returned credentials' user profile via `.actor`.
+
+> **Note:** When an actor token is provided, Auth0 does **not** issue a refresh token regardless of whether `offline_access` is in the requested scopes, so `credentials.refreshToken` will be `null` in this flow.
+
+> ⚠️ **Web requires `auth0-spa-js` >= 2.20.0.** Delegation/impersonation support (forwarding the actor token to `/oauth/token`) was added to the underlying [SPA SDK](https://github.com/auth0/auth0-spa-js) in [2.20.0](https://github.com/auth0/auth0-spa-js/releases/tag/v2.20.0). With an older SDK the `actor` is accepted but silently dropped before the request, so the returned ID token has no `act` claim and `credentials.user.actor` is `null`. Make sure the `<script>` tag in your `index.html` references 2.20.0 or later.
```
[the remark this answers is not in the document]
utkrishtsahu: Fixed — the blank line between the refresh-token note and the ⚠️ Web requires auth0-spa-js >= 2.20.0 warning is now a > continuation, so both paragraphs are one contiguous blockquote and MD028 passes.

auth0_flutter/lib/auth0_flutter.dart
```diff
@@ -3,14 +3,14 @@ import 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interfac
...
 import 'src/mobile/credentials_manager.dart';
-import 'src/mobile/mfa_api.dart';
 import 'src/mobile/my_account_api.dart';
 import 'src/mobile/passwordless_api.dart';
 import 'src/mobile/web_authentication.dart';
 import 'src/version.dart';
 
 export 'package:auth0_flutter_platform_interface/auth0_flutter_platform_interface.dart'
     show
+        ActorToken,
```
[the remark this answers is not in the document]
utkrishtsahu: Added — test/mobile/authentication_api_test.dart now has a "CTE delegation types are exported from the package barrel" group with a UserActor smoke test (nested actor + extraClaims) next to the existing ActorToken one, so a future show-list regression is caught.

auth0_flutter/lib/src/web/extensions/user_profile_extension.dart
```diff
@@ -118,6 +124,7 @@ class PublicClaims {
   static const phoneNumberVerified = 'phone_number_verified';
   static const address = 'address';
   static const updatedAt = 'updated_at';
+  static const act = 'act';
```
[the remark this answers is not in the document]
utkrishtsahu: Intentional — toList() feeds the "claims to strip from customClaims" set, and act is already added to that removal set explicitly (...PublicClaims.toList(), PublicClaims.act) and then parsed into the dedicated .actor field. Adding it to toList() would be redundant, so leaving it out keeps act routed to actor rather than generic claim handling.

pmathew92 (approved):

kishore7snehil (approved):
LGTM!

PR #892, opened 2026-07-10
refactor: Move darwin native sources into a canonical Sources/ tree

NandanPrabhu:
### 📋 Changes

Extracted from #804. Relocates `auth0_flutter/darwin/Classes/*.swift` into `auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/`, turning the old `Classes/` location — and the `ios/`/`macos/` CocoaPods directories — into symlinks generated by `scripts/generate-symlinks.sh`. This gives CocoaPods (and a future SPM target) a single canonical source location instead of duplicated/symlinked per-platform trees.

Also renames `SwiftAuth0FlutterPlugin` → `Auth0FlutterPlugin` and drops the Objective-C shim (`Auth0FlutterPlugin.h`/`.m`), since the class name can now match the `pluginClass` declared in `pubspec.yaml` directly without an ObjC bridge.

**This is a pure move + rename, no behavior change for CocoaPods consumers:**
- Every relocated file is byte-identical to its previous location (verified file-by-file against `main`).
- `darwin/auth0_flutter.podspec`'s `source_files` is updated to point at the new path; `ios/` and `macos/` podspecs are unchanged (they still glob `Classes/**/*`, which is now populated by regenerated symlinks).
- `scripts/generate-symlinks.sh` is updated to generate `darwin/Classes/<rel> -> ../../auth0_flutter/Sources/auth0_flutter/<rel>` and `ios|macos/Classes/<rel> -> ../../../darwin/Classes/<rel>`, verifying every symlink resolves. Confirmed idempotent (re-running the script against its own output produces no diff), so the existing `Symlinks` CI job continues to pass.

No Dart files, `pubspec.yaml`, or CI workflows are touched.

### 📎 References

- Precursor to #804 (SPM support), which will rebase onto this once merged.

### 🎯 Testing

- Verified every file moved from `darwin/Classes/` is byte-identical to its `main` counterpart (2 intentional exceptions: the plugin class rename and dropping the ObjC shim).
- `pod ipc spec auth0_flutter/darwin/auth0_flutter.podspec` validates the podspec and resolves `source_files` to the 64 real Swift files under the new `Sources/` path.
- Re-ran `scripts/generate-symlinks.sh` after committing; zero diff, confirming CI's `Symlinks` check will pass.
- `plutil -lint` on the updated example iOS `project.pbxproj` after the test-file rename.


NandanPrabhu:
Closing and reopening as a fresh PR to keep this cascading pair's PR numbers/descriptions in sync (see the SPM PR it now precedes: #893).

PR #893, opened 2026-07-10
feat: Swift Package Manager dependency support for Flutter iOS/macOS SDK

NandanPrabhu:
### 📋 Changes

Replaces #804, split into a cascading pair of PRs so the pure file-move (#894) can be reviewed separately from the actual SPM work. **Stacked on #894** — this PR's diff is scoped to just the SPM-specific delta; once #894 merges into `main`, this PR will retarget to `main` automatically and the diff will shrink to what's shown below.

Adds **Swift Package Manager (SPM)** support to the iOS/macOS (`darwin`) native plugin, addressing #793. **CocoaPods support is retained** as a fallback, so this is non-breaking for existing consumers.

**SPM packaging**
- Adds `auth0_flutter/darwin/auth0_flutter/Package.swift` declaring the `auth0_flutter` SPM target (iOS 14 / macOS 11), depending on Auth0.swift 2.24.1, JWTDecode.swift 3.3.0, and SimpleKeychain 1.3.0 — the same versions already pinned in the three podspecs.
- `PASSKEYS_PLATFORM` Swift compilation condition is set via `swiftSettings` in `Package.swift`, mirroring the podspecs' `SWIFT_ACTIVE_COMPILATION_CONDITIONS`.
- `pubspec.yaml`'s `sharedDarwinSource: true` (already on `main`) is what tells Flutter to resolve iOS/macOS from the single `Sources/auth0_flutter/` tree that #894 establishes.

**Test/source SPM-compatibility fixes**
- Adds explicit `Flutter`/`UIKit` imports to test files and `WebAuthModels.swift` that use `FlutterError`/`UIModalPresentationStyle` without importing them directly — needed under SPM, where these types aren't transitively visible from `@testable import auth0_flutter` the way CocoaPods' module map provides them.
- Switches CredentialsManager test fixtures from hand-built `NSKeyedArchiver` payloads to `credentialsManager.store(credentials:)`, exercising the real storage path instead of duplicating its serialization format in test code.

**CI**
- `setup-darwin`: enables SPM (`flutter config --enable-swift-package-manager`) before `flutter pub get`, then patches the Flutter-generated `FlutterGeneratedPluginSwiftPackage/Package.swift` deployment targets to 14.0 (iOS) / 11.0 (macOS) since Flutter's defaults are lower than this plugin's minimum.
- `setup-publish`: also enables SPM so publish-time `flutter pub get` resolves consistently.
- Drops the Ruby/`slather` toolchain from CI; iOS coverage is now `.xcresult` → Cobertura XML via `xcresultparser` (pinned to `2.0.1`) uploaded to Codecov.
- Removes the now-unused `example/ios/Gemfile(.lock)`.

**Example app**
- Podfiles: drop `use_modular_headers!` and the `RunnerUITests` CocoaPods target (no longer needed once the plugin builds under SPM); `RunnerUITests` keeps its Xcode target.
- macOS `AppDelegate`: `@NSApplicationMain` → `@main`, adds `applicationSupportsSecureRestorableState` for current Xcode/macOS AppKit expectations.
- Adds the generated `Package.resolved`/`.xcscheme` files and pbxproj SPM package-reference entries produced by resolving the plugin via SPM.

### 📎 References

- Fixes #793
- Stacked on #894
- Flutter SPM for plugin authors: https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors

### Review feedback from #804 addressed here

- ~~Podspec version (2.1.0) lagging pubspec.yaml (2.2.0)~~ — versions now tracked consistently; podspecs pin Auth0.swift 2.24.1 to match `Package.swift`.
- ~~Case-sensitive `inputs.platform` comparison in `setup-darwin` skips the deployment-target patch for lowercase `ios`/`macos`~~ — both `Fix … deployment target for SPM` steps now gate on the normalized `steps.lowercase-platform.outputs.platform`.
- ~~`generate-symlinks.sh` `../` depth miscalculation leaving ~185 dangling symlinks~~ — the symlink-generation logic itself is now owned by #894 (the move PR), which includes a post-generation resolution check.
- ~~`public_header_files`/`.{h,m}` glob entries pointing at a nonexistent `include/` dir~~ — the ObjC shim removal is now owned by #894; this PR's podspec only globs `**/*.swift`.
- ~~`xcresultparser` installed unpinned from a third-party tap~~ — pinned to `a7ex/homebrew-formulae/xcresultparser@2.0.1`.
- ~~Uncommented `.env` pubspec asset turning a missing `.env` into a hard build failure~~ — reverted; stays commented out.
- ~~SPM `.upToNextMajor` vs. podspecs' exact pin could drift~~ — both now pin Auth0.swift `2.24.1` exactly.

### Found while re-splitting the diff

- The example iOS `AppDelegate.swift` had lost `import Auth0` and the `application(_:open:options:)` override calling `WebAuthentication.resume(with: url)` — without it, the OAuth redirect callback wouldn't complete login. Restored to match `main`.
- `example/ios/Podfile`'s commented `# platform :ios, '13.0'` didn't match the actual `14.0` deployment target set in the Xcode project; corrected the comment to `14.0`.

### 🎯 Testing

- iOS and macOS native unit tests pass in CI via `xcodebuild` against `Runner.xcworkspace` (Xcode 26.2); Android and Windows unit suites unchanged.
- `pod ipc spec` and `swift package dump-package` both validate cleanly against the updated podspecs/`Package.swift`.
- Manual verification: from `auth0_flutter/example`, run `flutter config --enable-swift-package-manager` then `flutter run` on an iOS simulator or device; exercise login/logout, credentials manager, and DPoP flows.
- CocoaPods fallback can be verified by running `flutter config --no-enable-swift-package-manager` and building the example app.

## Summary by CodeRabbit

* **New Features**
  * Added Swift Package Manager support for iOS and macOS builds.
  * Added a Swift package manifest targeting iOS 14.0 and macOS 11.0.
  * Swift Package Manager now resolves native dependencies automatically when enabled.

* **Bug Fixes**
  * Improved coverage report generation and artifact upload in iOS CI workflows.

* **Documentation**
  * Added setup guidance for Swift Package Manager and CocoaPods integration.


auth0_flutter/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -818,7 +767,7 @@
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = Runner/RunnerDebug.entitlements;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 79ZXKC573P;
```
utkrishtsahu: 798, 827, 853, 878, 907, 933, 1064 in these lines too DEVELOPMENT_TEAM = 79ZXKC573P is hardcoded is it needed here as its team identifier can you please check.

auth0_flutter/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -70,6 +88,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
+      enableGPUValidationMode = "1"
```
utkrishtsahu: Xcode noise, unrelated to SPM. Drop it.

auth0_flutter/darwin/auth0_flutter/Package.swift
```diff
@@ -0,0 +1,30 @@
...
+import PackageDescription
+
+let package = Package(
+    name: "auth0_flutter",
+    platforms: [
+        .iOS("14.0"),
+        .macOS("11.0"),
+    ],
+    products: [
+        .library(name: "auth0-flutter", targets: ["auth0_flutter"]),
```
utkrishtsahu, line 11: Product is auth0-flutter (hyphen) while the target/package are auth0_flutter (underscore). Is the hyphen what Flutter's generated package expects? A mismatch here fails silently by falling back to CocoaPods — CI won't catch it. Please confirm.

.github/workflows/main.yml
```diff
@@ -151,15 +149,21 @@ jobs:
...
-            - name: Convert coverage report
+            # Codecov cannot parse a raw .xcresult bundle, so convert it to
+            # Cobertura XML first. xcresultparser is Ruby-free, keeping the SPM
+            # migration's goal of dropping the Ruby/slather toolchain.
+            - name: Convert xcresult to Cobertura coverage
               working-directory: auth0_flutter/example/ios
-              run: bundle exec slather coverage -x --scheme Runner Runner.xcodeproj
+              run: |
+                brew install a7ex/homebrew-formulae/xcresultparser@2.0.1
+                xcresultparser --output-format cobertura unit-tests.xcresult > coverage.xml
```
utkrishtsahu, line 159: xcresultparser ... unit-tests.xcresult hardcodes the bundle name. Does the test step actually emit unit-tests.xcresult? If not, tests pass but coverage fails — confusing to debug. (Also brew install runs every CI run; pinned so it's safe, just noting the cost.)

utkrishtsahu (changes requested):

NandanPrabhu:
@coderabbitai full review

auth0_flutter/example/ios/Runner.xcodeproj/project.pbxproj
NandanPrabhu: Good catch — this was leaked from my local Apple Developer Team ID when I regenerated the pbxproj during manual verification (running the app in Xcode locally). `main` has `DEVELOPMENT_TEAM = ""` in all 8 of these configs; this branch had accidentally overwritten them with my personal team ID.

Fixed in df00f11d — reverted all 8 occurrences back to `""`.

auth0_flutter/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
NandanPrabhu: Confirmed and agreed — `enableGPUValidationMode = "1"` is an Xcode-inserted local debug-run preference from when I ran/debugged the macOS app in Xcode during verification, not something SPM support requires.

Fixed in df00f11d — dropped the line.

auth0_flutter/darwin/auth0_flutter/Package.swift
NandanPrabhu: Verified directly against Flutter's own generated wrapper: `example/ios/Flutter/ephemeral/Packages/FlutterGeneratedPluginSwiftPackage/Package.swift` declares `.product(name: "auth0-flutter", package: "auth0_flutter")` — Flutter's tooling itself expects the **package** referenced by its underscore name but the **product** it links by the hyphenated name. This matches what's declared here.

This isn't a silent-fallback risk: if the product name didn't match, SPM resolution would fail to build outright rather than quietly falling back to CocoaPods. Confirmed working by running `xcodebuild test` with SPM enabled (322/322 passing) and reproduced green in the actual CI run for this PR.

.github/workflows/main.yml
NandanPrabhu: Yes — checked `.github/actions/unit-tests-darwin/action.yml`: `xcodebuild test ... -resultBundlePath unit-tests.xcresult` produces exactly that filename, matching what `xcresultparser` consumes.

Also confirmed in the actual CI run for this PR (run 29225262671) — the "Convert xcresult to Cobertura coverage" and "Upload coverage report" steps both succeeded, so this isn't just a naming assumption, it's proven working end-to-end.

On the `brew install` cost: agreed it's not cached across runs. It's pinned to `2.0.1` for reproducibility and only adds a few seconds, so it didn't seem worth the added cache complexity — happy to revisit if it becomes a real time sink.

.github/actions/setup-darwin/action.yml
```diff
@@ -75,6 +67,28 @@ runs:
...
+      if: steps.lowercase-platform.outputs.platform == 'ios'
+      run: |
+        PACKAGE_FILE="auth0_flutter/example/ios/Flutter/ephemeral/Packages/FlutterGeneratedPluginSwiftPackage/Package.swift"
+        if [ ! -f "$PACKAGE_FILE" ]; then
+          echo "::error::FlutterGeneratedPluginSwiftPackage/Package.swift not found at expected path"
+          exit 1
+        fi
+        sed -i '' 's/\.iOS(".*")/.iOS("14.0")/g' "$PACKAGE_FILE"
+      shell: bash
+
```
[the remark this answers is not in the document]
NandanPrabhu: Fair point on robustness, though low risk in practice — each generated `Package.swift` only has one `.iOS("...")`/`.macOS("...")` line, and I've run this exact `sed` against real Flutter-generated files (both locally and in CI, run 29225262671) with correct results every time.

That said, `[^"]*` is strictly safer with no downside, so happy to take this if you want it tightened up — just flag it and I'll push the change.

.github/workflows/main.yml
```diff
@@ -151,15 +149,21 @@ jobs:
...
+              shell: bash
```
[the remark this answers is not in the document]
NandanPrabhu: This is disproven by the actual CI run for this PR — checked run 29225262671, the "Convert xcresult to Cobertura coverage" step (which runs exactly `brew install a7ex/homebrew-formulae/xcresultparser@2.0.1`) completed successfully. Homebrew's versioned-formula syntax (`tap/formula@version`) does resolve against this tap's aliasing, contrary to the analysis above. No change needed — this is proven working, not theoretical.

auth0_flutter/darwin/auth0_flutter/Package.swift
```diff
@@ -0,0 +1,30 @@
...
+    ],
+    dependencies: [
+        .package(url: "https://github.com/auth0/Auth0.swift", exact: "2.24.1"),
+        .package(url: "https://github.com/auth0/JWTDecode.swift", exact: "3.3.0"),
+        .package(url: "https://github.com/auth0/SimpleKeychain", exact: "1.3.0"),
+    ],
```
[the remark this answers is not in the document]
NandanPrabhu: This was a deliberate, previously-made decision, not an oversight — see e81a3572 ("Pins Auth0.swift to exact 2.22.0 for both SPM and CocoaPods"): exact pins keep SPM and CocoaPods consumers resolving to the same Auth0.swift release and match the version already locked into the checked-in `Package.resolved` files under `example/ios`/`example/macos`. Switching to `.upToNextMajor`/`~>` ranges would reintroduce the exact SPM/CocoaPods version drift that commit fixed. Keeping exact pins.

auth0_flutter/example/macos/Podfile
```diff
@@ -28,18 +28,11 @@ flutter_macos_podfile_setup
...
   use_frameworks!
-  use_modular_headers!
 
   flutter_install_all_macos_pods File.dirname(File.realpath(__FILE__))
-
   target 'RunnerTests' do
     inherit! :search_paths
   end
-
-  target 'RunnerUITests' do
```
sanchitmehtagit, line 39: is this intended?

auth0_flutter/darwin/auth0_flutter.podspec
```diff
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.23.0'
+  s.dependency 'Auth0', '2.24.1'
```
sanchitmehtagit: we should not increase swift SDK version in this PR

auth0_flutter/example/macos/Podfile
NandanPrabhu: Yes, intentional. RunnerUITests no longer needs its own CocoaPods target — it's an Xcode-only UI test target with zero references to any Pods framework (verified: `grep -n "Pods.*RunnerUITests\|RunnerUITests.*Pods" Runner.xcodeproj/project.pbxproj` returns nothing). Since the plugin now builds via SPM, the extra Podfile block and `use_modular_headers!` were only there to support the old CocoaPods-based RunnerUITests target and are no longer needed.

Verified locally:
```
xcodebuild build -workspace macos/Runner.xcworkspace -scheme Runner -destination 'platform=macOS' -configuration Debug
# ** BUILD SUCCEEDED **

xcodebuild build-for-testing -workspace macos/Runner.xcworkspace -scheme Runner -destination 'platform=macOS' -configuration Debug -only-testing:RunnerUITests
# ** TEST BUILD SUCCEEDED **
```

auth0_flutter/darwin/auth0_flutter.podspec
NandanPrabhu: Good catch, agreed — reverted the Auth0.swift pin back to 2.23.0 (the version already on main) across all three podspecs, `Package.swift`, and the checked-in `Package.resolved` files, in bc2ddee8. 2.24.1 only contained a watchOS-specific bugfix unrelated to SPM, so bumping it here was scope creep from an earlier iteration of this branch. Verified the reverted build/tests pass on both iOS and macOS.

utkrishtsahu (approved):

PR #894, opened 2026-07-10
refactor: Move darwin native sources into a canonical Sources/ tree

NandanPrabhu:
### 📋 Changes

First of a cascading pair of PRs replacing #804 (closed — see [comment](https://github.com/auth0/auth0-flutter/pull/804#issuecomment-4933489094) for why it was split). Relocates `auth0_flutter/darwin/Classes/*.swift` into `auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/`, turning the old `Classes/` location — and the `ios/`/`macos/` CocoaPods directories — into symlinks generated by `scripts/generate-symlinks.sh`. This gives CocoaPods (and #893's SPM target) a single canonical source location instead of duplicated/symlinked per-platform trees.

Also renames `SwiftAuth0FlutterPlugin` → `Auth0FlutterPlugin` and drops the Objective-C shim (`Auth0FlutterPlugin.h`/`.m`), since the class name can now match the `pluginClass` declared in `pubspec.yaml` directly without an ObjC bridge.

**This is a pure move + rename for production code — no behavior change for CocoaPods consumers.** One test-only fixture does change behavior (see below):
- Every relocated production file is byte-identical to its previous location (verified file-by-file against `main`).
- `darwin/auth0_flutter.podspec`'s `source_files` is updated to point at the new path; `ios/` and `macos/` podspecs are unchanged (they still glob `Classes/**/*`, which is now populated by regenerated symlinks).
- `scripts/generate-symlinks.sh` is updated to generate `darwin/Classes/<rel> -> ../../auth0_flutter/Sources/auth0_flutter/<rel>` and `ios|macos/Classes/<rel> -> ../../../darwin/Classes/<rel>`, verifying every symlink resolves. Confirmed idempotent (re-running the script against its own output produces no diff), so the existing `Symlinks` CI job continues to pass.
- `Tests/CredentialsManager/CredentialsManagerSpies.swift`'s `SpyCredentialsStorage.setEntry` now writes its argument into `getEntryReturnValue`, making the spy round-trip like the real Keychain-backed storage. This is what lets `CredentialsManagerGetMethodHandlerTests`/`CredentialsManagerHasValidMethodHandlerTests`/etc. use `credentialsManager.store(credentials:)` (which calls `setEntry` then reads it back via `getEntry`) instead of hand-building `NSKeyedArchiver` payloads that duplicated the real storage's serialization format. Checked every call site — no test sets `getEntryReturnValue` and later asserts it stays at that value after a `setEntry` call, so nothing is clobbered by the change.

No Dart files, `pubspec.yaml`, or CI workflows are touched.

### 📎 References

- Precursor to #893 (SPM support), which is stacked on this branch and will retarget to `main` once this merges.

### 🎯 Testing

- Verified every file moved from `darwin/Classes/` is byte-identical to its `main` counterpart (2 intentional exceptions: the plugin class rename and dropping the ObjC shim).
- `pod ipc spec auth0_flutter/darwin/auth0_flutter.podspec` validates the podspec and resolves `source_files` to the 64 real Swift files under the new `Sources/` path.
- Re-ran `scripts/generate-symlinks.sh` after committing; zero diff, confirming CI's `Symlinks` check will pass.
- `plutil -lint` on the updated example iOS `project.pbxproj` after the test-file rename.


auth0_flutter/example/ios/Tests/CredentialsManager/CredentialsManagerSpies.swift
```diff
@@ -18,6 +19,7 @@ class SpyCredentialsStorage: CredentialsStorage {
 
     func setEntry(_ data: Data, forKey key: String) -> Bool {
         self.calledSetEntry = true
+        self.getEntryReturnValue = data
```
utkrishtsahu, line 22: setEntry now writes into getEntryReturnValue, turning the spy into a real read-back store. Small line, but it's behavior — not a move. Call it out in the description.

auth0_flutter/example/ios/Tests/Mocks.swift
```diff
@@ -113,7 +113,8 @@ class SpyPluginRegistrar: NSObject, FlutterPluginRegistrar {
     }
 
     func addApplicationDelegate(_ delegate: FlutterPlugin) {}
-    func addSceneDelegate(_ delegate: any FlutterSceneLifeCycleDelegate) {}
+
+    func addSceneDelegate(_ delegate: FlutterSceneLifeCycleDelegate) {}
```
utkrishtsahu, line 117: dropped any from addSceneDelegate(_ delegate: FlutterSceneLifeCycleDelegate). Does that still match the protocol on the Flutter version SPM resolves? If the protocol still declares any, this won't compile in Swift 6 mode. Please confirm.

auth0_flutter/darwin/auth0_flutter.podspec
utkrishtsahu, line 22: This still pins Auth0 2.23.0 while #893 bumps it to 2.24.1 here and in the ios/macos podspecs by hand. please check the sync

utkrishtsahu (changes requested):

auth0_flutter/darwin/auth0_flutter.podspec
NandanPrabhu: This we can ignore because either we have version bump changes here or the other PR we will have version bump anyways

NandanPrabhu:
@coderabbitai full review

auth0_flutter/example/ios/Tests/CredentialsManager/CredentialsManagerSpies.swift
NandanPrabhu: Confirmed intentional, not a bug. `setEntry` writing back into `getEntryReturnValue` makes the spy round-trip like the real Keychain storage does — it's what let us swap the CredentialsManager test fixtures from hand-built `NSKeyedArchiver` payloads to `credentialsManager.store(credentials:)` (which calls `setEntry` then reads it back via `getEntry`), exercising the real storage path instead of duplicating its serialization format in test code.

Checked every call site across `CredentialsManagerSaveMethodHandlerTests`, `CredentialsManagerRenewMethodHandlerTests`, etc. — none set `getEntryReturnValue` and then assert it stays at that value after a later `setEntry` call, so nothing gets clobbered.

Agreed this is a behavior change worth calling out explicitly — added a line to the PR description.

auth0_flutter/example/ios/Tests/Mocks.swift
NandanPrabhu: Checked against the actual Flutter engine header (`FlutterPlugin.h`) shipped in this project's `Flutter.framework` — the protocol signature is unchanged. This project builds with `SWIFT_VERSION = 5.0` (not Swift 6 mode), where `any` before a protocol-typed parameter is optional, not required.

Verified directly: ran the full `xcodebuild test` suite (322 tests) against this file under both SPM and CocoaPods dependency modes — compiles and passes both ways. No compile risk here.

auth0_flutter/darwin/auth0_flutter.podspec
NandanPrabhu: Confirmed resolved — checked all three podspecs (`darwin`, `ios`, `macos`) on both #894 and #893, all three pin `2.24.1` on both branches at the current tip. This comment was against an earlier commit; the version bump landed in a later commit already included in the current diff.

NandanPrabhu:
@coderabbitai full review

utkrishtsahu (approved):

utkrishtsahu (approved):

PR #904, opened 2026-07-23
feat: surface IPSIE session_expiry claim across Android, iOS and web

utkrishtsahu:

### 📋 Changes

Adds support for the IPSIE `session_expiry` claim — an absolute Unix-seconds ceiling asserted by an upstream IdP (over an enterprise connection with **"Use ID Token for Session Expiry"** / `id_token_session_expiry_supported: true` enabled). The claim acts as a hard ceiling on the local session: the app session can never outlive the upstream IdP session.

> ⚠️ **Early Access:** this feature is Early Access; the API surface (`sessionExpiry` field and `isSessionExpired`) may change before GA.

Enforcement itself lives in the native SDKs; this change bumps to the versions that ship it, surfaces the value through the Dart `Credentials` model, and carries it across the platform-channel boundary on all three platforms.

**Public API**
- **`Credentials`** (`auth0_flutter_platform_interface`): new optional, nullable `DateTime? sessionExpiry` field, wired through the constructor, `fromMap`, and `toMap`. Additive and non-breaking — `null` means "no ceiling" and is never treated as expired. `fromMap` validates the value with an `is String` check (throws `FormatException` on unexpected types); `toMap` omits the key entirely when there is no ceiling, matching the native serializers.
  ```dart
  final credentials = await auth0.credentialsManager.credentials();
  final sessionExpiry = credentials.sessionExpiry; // null when the claim is absent
  if (sessionExpiry != null) {
    print('Upstream IdP session ends at: $sessionExpiry');
  }
  ```
- **`CredentialsManagerException.isSessionExpired`**: new getter. Once the ceiling passes, the native SDK raises a dedicated session-expired error (Android `SESSION_EXPIRED`, iOS `.sessionExpired`) — distinct from "no credentials". Apps branch on `isSessionExpired` to trigger a fresh login:
  ```dart
  try {
    final credentials = await auth0.credentialsManager.credentials();
  } on CredentialsManagerException catch (e) {
    if (e.isSessionExpired) {
      // Upstream IdP session_expiry ceiling reached — re-authenticate.
    }
  }
  ```

**Serialization** (reads the value the credentials manager actually enforces, not the raw claim)
- **Darwin (iOS/macOS):** `Extensions.swift` reads the native `Credentials.sessionExpiresAt` in `asDictionary()` (pinned-at-login value, with the native millisecond-vs-seconds range check); `sessionExpiry` case in `Properties.swift`; `.sessionExpired` mapped to `SESSION_EXPIRED` in `CredentialsManagerExtensions.swift`.
- **Android:** shared `Credentials.toMap()` extension (`CredentialsExtensions.kt`) reads the native `credentials.sessionExpiresAt`, used by the login, get, and renew handlers (deduplicates three copy-pasted maps). Core-library desugaring enabled for `java.time` on `minSdk 21`.
- **Web:** `credentials_extension.dart` decodes the claim from the ID token, applying the same `(0, 10_000_000_000)` bound as native so a millisecond value reads as "no ceiling".

**Dependency bumps** (minor line that introduced the feature, not the next major)
- Auth0.swift → **2.24.1** (all podspecs + `Package.swift`)
- auth0-spa-js → **2.22.0** (example `web/index.html`)
- Auth0.Android already at 3.21.0 (≥ 3.20.0) — no change needed

**Docs:** `README.md` note + new "Session expiry from an upstream IdP" section in `EXAMPLES.md` (covers `isSessionExpired` handling, the seconds-not-milliseconds gotcha, the ~30s clock-skew leeway, EA status, and links the IPSIE SL1 spec).

### 📎 References

SDK-10259

### 🎯 Testing

**Automated (added in this PR)**
- **Platform interface** (`credentials_test.dart`): `sessionExpiry` present/absent via `fromMap`; ISO-8601 UTC output and key-omission via `toMap`; `FormatException` on a non-String value.
- **Web** (`credentials_extension_test.dart`): claim decoded from a signed test token; `null` when absent; `null` for a millisecond-valued claim (bound check).
- **Android** (`GetCredentialsRequestHandlerTest` / `RenewCredentialsRequestHandlerTest` / `LoginWebAuthRequestHandlerTest`): claim presence/absence across the get, renew, **and login** paths, and **preservation across renew** (a refresh must not extend the ceiling); `JwtTestUtils` extended with numeric claims.
- **iOS/macOS** (`ExtensionsTests`): `asDictionary` present/absent cases; shared test token uses a valid in-range seconds ceiling.

Run:
```bash
# platform interface
cd auth0_flutter_platform_interface && flutter test test/credentials_test.dart
# web
cd auth0_flutter && flutter test test/web/extensions/credentials_extension_test.dart --platform chrome
# android
cd auth0_flutter/example/android && ./gradlew :auth0_flutter:testDebugUnitTest
```

**Manual (end-to-end)** — verified against a live tenant on **Android** and **Web** (iOS shares the same Dart layer + the confirmed `.sessionExpired` mapping):
1. Configure an enterprise connection with an upstream IdP and enable **"Use ID Token for Session Expiry"** (or inject the claim via a Post-Login Action for testing).
2. Log in; confirm `credentials.sessionExpiry` is non-null and matches the `session_expiry` claim.
3. Call `renewCredentials()` and confirm the ceiling is unchanged (not pushed later).
4. Wait for the ceiling to pass, then call `credentials()` — confirm it raises `CredentialsManagerException` with `isSessionExpired`, and that a fresh login succeeds.


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsExtensions.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+/**
+ * Reads the `session_expiry` claim from the decoded ID token (exposed via the
+ * user profile's extra info) and converts the Unix-seconds value to an ISO-8601
+ * UTC string. Returns `null` when the claim is absent or not numeric.
+ */
+private fun Credentials.sessionExpiryIso(): String? {
+    val claim = user.getExtraInfo()["session_expiry"] as? Number ?: return null
```
NandanPrabhu: credentials exposes sessionExpiresAt can we not use it directly?

auth0_flutter/EXAMPLES.md
```diff
@@ -700,6 +701,43 @@ final credentials = await auth0.credentialsManager.credentials();
...
+
+// null when the connection option is not enabled, or the session predates the
+// feature. A null value means "no upstream ceiling".
+final sessionExpiry = credentials.sessionExpiry;
+if (sessionExpiry != null) {
+  print('Upstream IdP session ends at: $sessionExpiry');
+}
+```
+
+> ⚠️ **Upgrade note:** once this feature is enabled on your connection, `credentials()` can raise a "no credentials" error for a user who was previously logged in, when the ceiling is reached. If your app assumed `credentials()` always resolves after login, make sure it handles this case (it already does if you catch `CredentialsManagerException` as shown above).
```
NandanPrabhu: should we mention IPSIE as EA?

auth0_flutter/EXAMPLES.md
```diff
@@ -700,6 +701,43 @@ final credentials = await auth0.credentialsManager.credentials();
...
+When a user authenticates through an upstream Identity Provider (for example Okta) over an enterprise connection that has **"Use ID Token for Session Expiry"** enabled (`id_token_session_expiry_supported: true`), Auth0 adds a `session_expiry` claim to the ID token. This claim is an absolute point in time (a Unix-seconds timestamp) that represents when the upstream IdP session ends, and it acts as a **hard ceiling** on the local session: the app session can never outlive the upstream IdP session.
+
+**Enforcement is transparent — you do not need to write any new code.** Once the ceiling passes, the underlying native SDK treats the stored credentials as expired and skips refresh-token renewal, so `credentials()` raises a "no credentials" error exactly as it would for a logged-out user. Your existing "needs login" handling then triggers a fresh login:
+
+```dart
+try {
+  final credentials = await auth0.credentialsManager.credentials();
+  // Use the credentials
+} on CredentialsManagerException catch (e) {
+  if (e.isNoCredentialsFound) {
```
kishore7snehil: The guide tells developers to branch on `isNoCredentialsFound` when the ceiling is reached, but that check doesn't actually match the ceiling error on mobile. On Android the native SDK raises a distinct `SESSION_EXPIRED` error whose message ("The session has reached the session_expiry ceiling…") comes through as the exception code, and on iOS it raises its own session-expired case that the platform channel maps to `UNKNOWN`. `isNoCredentialsFound` only matches `NO_CREDENTIALS`, so on both platforms this catch quietly falls through and the app never runs its re-login path, which is the whole point of the feature.

Can we surface the session-expired case as its own signal (say an `isSessionExpired` on the exception) and have the example branch on that? Right now there's nothing in the Dart layer that maps it.

auth0_flutter/lib/src/web/extensions/credentials_extension.dart
```diff
@@ -13,13 +13,25 @@ extension CredentialsExtension on Credentials {
     final expiresAt = DateTime.now().add(Duration(seconds: expiresIn));
     final claims = JWT.decode(webCredentials.id_token);
     final user = UserProfileExtension.fromClaims(claims);
+    // IPSIE session_expiry: an absolute Unix-seconds ceiling asserted by the
+    // upstream IdP. auth0-spa-js (>= 2.22.0) enforces it on silent renewal;
+    // here we surface the value so app code can read
+    // `credentials.sessionExpiry`.
+    final sessionExpiryClaim = claims['session_expiry'];
+    final sessionExpiry = sessionExpiryClaim is num
```
kishore7snehil: There's no upper-bound check on the claim here before turning it into a date. The native layers treat a `session_expiry` at or above 10,000,000,000 as "no ceiling" (the value is almost certainly milliseconds, since a real seconds value that large is year 2286+), so enforcement returns null for it. Here we'd still build a far-future `DateTime` and hand it back, so `credentials.sessionExpiry` disagrees with what's actually enforced. The claim is author-set in a Post-Login Action, so a forgotten seconds conversion is a real case.

Shall we apply the same bound the native SDKs use so a millisecond value reads as "no ceiling" here too? The same gap exists on the Android and iOS decoders.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsExtensions.kt
kishore7snehil: This reads `session_expiry` off the current ID token's claims, but the native `Credentials` already exposes `sessionExpiresAt`, and that property returns the value pinned at the initial login, which is the value the credentials manager actually enforces. The two can diverge: after a refresh whose ID token omits the claim (or re-emits a different one), the pinned ceiling still enforces while the live token no longer carries it, so what we surface here would drift from what the SDK is enforcing. It also means the ms-vs-seconds bound the native side applies is bypassed.

Can we read `sessionExpiresAt` directly instead of decoding the claim ourselves? That keeps the surfaced value in step with enforcement and picks up the native validation for free.

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/Extensions.swift
```diff
@@ -91,6 +91,14 @@ extension Credentials {
             CredentialsProperty.tokenType.rawValue: tokenType
         ]
         data[CredentialsProperty.refreshToken] = refreshToken
+        // IPSIE session_expiry: an absolute Unix-seconds ceiling asserted by the
+        // upstream IdP. Surfaced to the Dart layer as an ISO-8601 string so app
+        // code can read `credentials.sessionExpiry`; enforcement itself is done
+        // by the native CredentialsManager. Absent claim => omit (no ceiling).
+        if let sessionExpiry = (jwt.body["session_expiry"] as? NSNumber)?.doubleValue {
```
kishore7snehil: Same as on Android: this decodes the claim off the current ID token, but the native `Credentials` already exposes `sessionExpiresAt`, which reflects the value pinned at login (the one the CredentialsManager enforces) and applies the millisecond range check. Decoding the live token here can drift from the enforced ceiling after a refresh and skips that range check.

Can we read the native `sessionExpiresAt` here instead of parsing `jwt.body` ourselves?

auth0_flutter_platform_interface/lib/src/credentials.dart
```diff
@@ -48,6 +48,25 @@ class Credentials {
...
+  /// reached and will not renew them past it, regardless of [expiresAt].
+  ///
+  /// It is layered *on top of* the access-token [expiresAt] and any idle/
+  /// absolute timeouts — not a replacement for them.
+  ///
+  /// This is `null` when the claim is absent (the connection option is not
+  /// enabled, or the session predates the feature). A `null` value means
+  /// **no ceiling** and must never be treated as an already-expired session.
+  ///
+  /// [Read more about upstream IdP session expiry](https://auth0.com/docs).
```
kishore7snehil: This links to the docs root rather than a page about the feature.

auth0_flutter/example/ios/Tests/Mocks.swift
```diff
@@ -6,9 +6,10 @@ import Flutter
 import FlutterMacOS
 #endif
 
+
 let testIdToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJmb28iLCJuYW1lIjoiYmFyIiwiZW1haWwiOiJmb29AZXhhbXBsZS5"
     + "jb20iLCJlbWFpbF92ZXJpZmllZCI6dHJ1ZSwicGljdHVyZSI6Imh0dHBzOi8vZXhhbXBsZS5jb20vcGljdHVyZSIsInVwZGF0ZWRfYXQiOiIyMDI"
-    + "yLTA0LTE1VDAzOjE1OjUxLjc4N1oifQ.mFq-johzLTFQUAl9pjgQraTM6I8AGfcEcWBg0Ah2vss"
+    + "yLTA0LTE1VDAzOjE1OjUxLjc4N1oiLCJzZXNzaW9uX2V4cGlyeSI6MzI1MDM2ODAwMDB9.sig"
```
kishore7snehil: This adds `session_expiry` to the shared token, but there's no iOS test asserting the value actually lands in the credentials map, unlike the Android, web, and Dart sides which each got present/absent coverage. Worth a `toMap`/`asDictionary` case in the extensions tests so the iOS mapping is pinned too. (The value baked in here, `32503680000`, is also a 13-ish digit value that the native range check would reject as milliseconds, so it's not a great stand-in for a real ceiling.)

auth0_flutter/EXAMPLES.md
```diff
@@ -700,6 +701,43 @@ final credentials = await auth0.credentialsManager.credentials();
 
 > 💡 You do not need to call `credentialsManager.storeCredentials()` afterward. The Credentials Manager automatically persists the renewed credentials.
 
+### 🕒 Session expiry from an upstream IdP
```
kishore7snehil, line 704: Should this section carry an availability note? This SDK is targeted later than the ones that shipped the feature first, and the underlying native support went out as early access, so it may be worth telling developers where this stands rather than reading as fully GA.

kishore7snehil (changes requested):

auth0_flutter_platform_interface/lib/src/credentials.dart
```diff
@@ -78,5 +101,6 @@ class Credentials {
         'scopes': scopes.toList(),
         'userProfile': user.toMap(),
         'tokenType': tokenType,
+        'sessionExpiry': sessionExpiry?.toUtc().toIso8601String(),
```
kishore7snehil: `toMap` always writes the `sessionExpiry` key, so when it's null the map carries `sessionExpiry: null`, while the Android side omits the key entirely when there's no claim. `fromMap` reads both fine, so nothing breaks, but the two sides disagree on shape if anything ever checks for key presence.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
```diff
@@ -80,20 +80,7 @@ class LoginWebAuthRequestHandler(
...
-                        "accessToken" to credentials.accessToken,
-                        "idToken" to credentials.idToken,
-                        "refreshToken" to credentials.refreshToken,
-                        "userProfile" to credentials.user.toMap(),
-                        "expiresAt" to formattedDate,
-                        "scopes" to scopes,
-                        "tokenType" to credentials.type
-                    )
-                )
+                result.success(credentials.toMap())
```
kishore7snehil, line 83: This path now goes through the shared `toMap()`, but there's no test covering `sessionExpiry` on the web-auth login result the way `GetCredentials` and `RenewCredentials` got. Worth a present/absent case here too so the login path is pinned like the others.

kishore7snehil (changes requested):

auth0_flutter/EXAMPLES.md
```diff
@@ -700,6 +701,43 @@ final credentials = await auth0.credentialsManager.credentials();
...
+
+When a user authenticates through an upstream Identity Provider (for example Okta) over an enterprise connection that has **"Use ID Token for Session Expiry"** enabled (`id_token_session_expiry_supported: true`), Auth0 adds a `session_expiry` claim to the ID token. This claim is an absolute point in time (a Unix-seconds timestamp) that represents when the upstream IdP session ends, and it acts as a **hard ceiling** on the local session: the app session can never outlive the upstream IdP session.
```
kishore7snehil, line 708: This says the claim is a Unix-seconds timestamp, but doesn't call out the common mistake of emitting milliseconds. When the ceiling is set by a Post-Login Action, a developer who forgets to divide by 1000 produces a value that reads as tens of thousands of years out, which the platform treats as no ceiling, so enforcement silently switches off rather than erroring.

Shall we add a short warning that the claim is in seconds and a milliseconds value disables the ceiling? For the wording, the server-python guide already has one you could mirror: https://github.com/auth0/auth0-server-python/blob/2c7e3f80bf2702c3964eefda978b9ed4f55183e8/examples/RetrievingData.md?plain=1#L81-L82

auth0_flutter/EXAMPLES.md
```diff
@@ -700,6 +701,43 @@ final credentials = await auth0.credentialsManager.credentials();
...
+  if (e.isNoCredentialsFound) {
+    // Session ended (this includes the upstream IdP session_expiry ceiling
+    // being reached) — send the user through the login flow again.
+  }
+}
+```
+
+The ceiling is **layered on top of** the access-token expiry and any idle/absolute timeouts — it does not replace them. The session ends at whichever limit is reached first.
+
+You can also read the value for your own app logic (for example, to show a session countdown) via the new `sessionExpiry` field on `Credentials`:
```
kishore7snehil, line 730: Since this shows reading the value for a countdown, it's worth noting that enforcement applies a small leeway (about 30 seconds) for clock skew, so the session is treated as expired slightly before this exact timestamp. Otherwise someone building a countdown off this value may be surprised the session ends a bit early.

The server-python guide phrases this well if you want to mirror it: https://github.com/auth0/auth0-server-python/blob/2c7e3f80bf2702c3964eefda978b9ed4f55183e8/examples/RetrievingData.md?plain=1#L130

kishore7snehil (changes requested):

auth0_flutter/test/web/extensions/credentials_extension_test.dart
```diff
@@ -53,5 +62,32 @@ void main() {
       expect(result.refreshToken, refreshToken);
       expect(result.scopes, {...scope.splitBySingleSpace()});
     });
+
+    test('decodes sessionExpiry from the session_expiry claim', () {
```
kishore7snehil, line 66: The web decode here only has present and absent cases. The other SDKs  reject implausible values (a millisecond-scale timestamp, and out-of-range values generally), but there's no test for that on web, and right now the web decode doesn't apply that bound at all. If the decode grows the same range check the other layers use, we can add a test here covering a millisecond value, zero, negative, and a non-numeric claim, so the fail-open behavior is pinned.

auth0_flutter_platform_interface/lib/src/credentials.dart
```diff
@@ -48,6 +48,25 @@ class Credentials {
...
   final String tokenType;
 
+  /// The absolute date and time at which the upstream Identity Provider (IdP)
+  /// session expires, when the connection has **"Use ID Token for Session
+  /// Expiry"** enabled (`id_token_session_expiry_supported: true`).
+  ///
+  /// This is derived from the `session_expiry` claim (a Unix-seconds timestamp)
+  /// in the ID token and acts as a hard ceiling on the local session: the
+  /// underlying native SDK treats the credentials as expired once this time is
+  /// reached and will not renew them past it, regardless of [expiresAt].
```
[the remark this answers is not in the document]
utkrishtsahu: Done — reworded to "underlying platform SDK" and split native (CredentialsManager) vs web (auth0-spa-js on silent renewal) enforcement.

auth0_flutter_platform_interface/lib/src/credentials.dart
```diff
@@ -68,6 +88,9 @@ class Credentials {
         user: UserProfile.fromMap(Map<String, dynamic>.from(
             result['userProfile'] as Map<dynamic, dynamic>)),
         tokenType: result['tokenType'] as String,
+        sessionExpiry: result['sessionExpiry'] == null
+            ? null
+            : DateTime.parse(result['sessionExpiry'] as String).toUtc(),
```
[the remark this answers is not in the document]
utkrishtsahu: Done — added a _parseSessionExpiry helper that returns null for null, parses String→UTC, and throws FormatException for any other type. No more unchecked cast.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsExtensions.kt
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * The IPSIE `session_expiry` claim is an absolute Unix-seconds ceiling asserted
+ * by the upstream IdP. It is surfaced to the Dart layer as an ISO-8601 string
+ * (`sessionExpiry`) so app code can read `credentials.sessionExpiry`; the
+ * enforcement itself is performed by the native `SecureCredentialsManager`.
+ * When the claim is absent the key is omitted (meaning "no ceiling").
+ */
+fun Credentials.toMap(): Map<String, Any?> {
+    val scopes = this.scope?.split(" ") ?: listOf()
+    val formattedDate = this.expiresAt.toInstant().toString()
```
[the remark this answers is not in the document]
utkrishtsahu: Enabled coreLibraryDesugaring in the plugin's build.gradle (+ desugar_jdk_libs). Note the pre-existing expiresAt already used java.time; desugaring covers both.

auth0_flutter/lib/src/web/extensions/credentials_extension.dart
```diff
@@ -13,13 +13,25 @@ extension CredentialsExtension on Credentials {
...
+        ? DateTime.fromMillisecondsSinceEpoch(
+            sessionExpiryClaim.toInt() * 1000,
+            isUtc: true,
+          )
+        : null;
```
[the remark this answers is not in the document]
utkrishtsahu: Done — typed the claim as Object?; the is num guard still promotes safely.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsExtensions.kt
utkrishtsahu: Yes — switched to native credentials.sessionExpiresAt (pinned-at-login value the manager enforces) instead of decoding the claim. Verified on-device.

auth0_flutter/EXAMPLES.md
utkrishtsahu: Added an Early Access callout at the top of the section.

auth0_flutter/EXAMPLES.md
utkrishtsahu: Fixed — added isSessionExpired on CredentialsManagerException (maps Android SESSION_EXPIRED + iOS .sessionExpired) and updated the guide to branch on it. Verified end-to-end on Android: isSessionExpired=true, isNoCredentialsFound=false.

auth0_flutter/lib/src/web/extensions/credentials_extension.dart
utkrishtsahu: Added the same (0, 10_000_000_000) bound as native, so millisecond values read as "no ceiling". Test added.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/CredentialsExtensions.kt
utkrishtsahu: Switched to native sessionExpiresAt — reflects the pinned ceiling and picks up the native range check. Same fix as your iOS comment.

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/Extensions.swift
utkrishtsahu: Replaced the docs-root link with the IPSIE SL1 spec URL (matches how the native/RN SDKs reference it).

auth0_flutter_platform_interface/lib/src/credentials.dart
utkrishtsahu: Fixed — toMap now omits sessionExpiry when null, matching the native serializers. Test updated.

auth0_flutter/example/ios/Tests/Mocks.swift
utkrishtsahu: Changed the fixture to 4102444800 (year 2100, within the native bound) and added iOS asDictionary present/absent tests.

auth0_flutter/EXAMPLES.md
utkrishtsahu: Added an Early Access note to the section.

auth0_flutter_platform_interface/lib/src/credentials.dart
utkrishtsahu: Fixed in c83ef0e. toMap now only emits the key when a ceiling is present (if (sessionExpiry != null) 'sessionExpiry': ...), so it matches the native serializers instead of carrying sessionExpiry: null. Updated the test to assert the key is absent (containsKey('sessionExpiry') == false) rather than null.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/request_handlers/web_auth/LoginWebAuthRequestHandler.kt
utkrishtsahu: Added in c83ef0e — present/absent sessionExpiry cases on the web-auth login result in LoginWebAuthRequestHandlerTest, matching the GetCredentials/RenewCredentials coverage. Present asserts the ISO value; absent asserts the key is omitted.

auth0_flutter/EXAMPLES.md
utkrishtsahu: added a warning right after the claim is introduced: the value must be in seconds, a milliseconds value reads as tens of thousands of years out and is treated as no ceiling (enforcement silently off), with the Math.floor(Date.now() / 1000) reminder for Post-Login Actions.

auth0_flutter/EXAMPLES.md
utkrishtsahu: Added — noted that enforcement applies a ~30s negative leeway, so the session is treated as expired slightly before the timestamp; called out to account for it when building a countdown off sessionExpiry.

kishore7snehil (changes requested):

auth0_flutter/test/web/extensions/credentials_extension_test.dart
utkrishtsahu: Added.The web decode already applies the same (0, 10_000_000_000) bound as the native layers (added in this PR). Added tests pinning the fail-open behaviour: millisecond value, zero, negative, non-numeric, and the exclusive upper bound all decode to null, alongside the existing valid-seconds and absent cases. All pass.

kishore7snehil (approved):
LGTM!

NandanPrabhu (approved):

PR #909, opened 2026-07-28
feat: Upgrade Auth0.swift to v3.0.2

NandanPrabhu:
## Summary
- Upgraded Auth0.swift dependency from v2.x to v3.0.1 for iOS, macOS, tvOS, and watchOS
- Updated all Darwin plugin method handlers to handle Auth0 v3 API changes
- Updated test suite to match new Auth0.swift v3.0.1 signatures
- Removed deprecated APIs and adapted to new error handling patterns

### Dependency Updates
- Updated podspec files to use Auth0.swift 3.0.2
- Updated Swift Package Manager dependencies

### API Migration
- **Error Handling**: Migrated from `.webAuthError()` to `.authenticationError()` for WebAuth errors
- **Credentials Manager**: Updated credential manager API calls for v3 signatures
- **Auth API**: Adapted authentication API method handlers for new response types
- **MFA API**: Updated multifactor authentication handlers
- **Passkey Support**: Updated passkey credential exchange methods

### Test Suite Updates
- Updated test mocks and spies to match Auth0.swift v3.0.1 signatures
- Fixed error handling assertions in test cases
- Updated method signatures across all test files

## Test Plan
- [x] All existing unit tests pass
- [x] iOS example app runs successfully
- [x] Login/logout flow works
- [x] Credentials manager operations work
- [x] MFA flows work correctly
- [x] Passkey authentication works


auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/AuthAPI/AuthAPIHandler.swift
```diff
@@ -64,11 +64,21 @@ public class AuthAPIHandler: NSObject, FlutterPlugin {
         return client
     }
 
+    // `Authentication.login(withOTP:mfaToken:)` and `Authentication.multifactorChallenge(...)`
```
NandanPrabhu: we can remove these methods completely may be in another PR but right now ensuring no or min breaking change on dart side

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/CredentialsManager/CredentialsManagerSaveMethodHandler.swift
```diff
@@ -19,6 +19,11 @@ struct CredentialsManagerSaveMethodHandler: MethodHandler {
             return callback(FlutterError(from: .requiredArgumentMissing(Argument.credentials.rawValue)))
         }
 
-        callback(self.credentialsManager.store(credentials: credentials))
+        do {
+            try self.credentialsManager.store(credentials: credentials)
+            callback(true)
```
utkrishtsahu, line 24: store now throws a typed CredentialsManagerError, but this collapses it to a bare false — the caller can't tell why the save failed. Map the caught error to a FlutterError, the way ClearApiCredentialsMethodHandler does in this same PR. Right now the two credential handlers are inconsistent and the new throwing contract buys nothing.

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/CredentialsManager/CredentialsManagerClearMethodHandler.swift
```diff
@@ -10,6 +10,11 @@ struct CredentialsManagerClearMethodHandler: MethodHandler {
     let credentialsManager: CredentialsManager
 
     func handle(with arguments: [String: Any], callback: @escaping FlutterResult) {
-        callback(self.credentialsManager.clear())
+        do {
+            try self.credentialsManager.clear()
+            callback(true)
```
utkrishtsahu, line 15: Same issue: the thrown error is discarded. Either propagate FlutterError(from: CredentialsManagerError) or state why clear intentionally stays boolean-only.

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/AuthAPI/AuthAPIHandler.swift
```diff
@@ -64,11 +64,21 @@ public class AuthAPIHandler: NSObject, FlutterPlugin {
...
+    // were removed in Auth0.swift v3. These two method handlers use the dedicated `MFAClient`
+    // instead (see Auth0.swift's V3_MIGRATION_GUIDE.md), so they need an `MFAClient` rather than
+    // the `Authentication` client passed to every other handler here.
+    var mfaClientProvider: (_ account: Account) -> MFAClient = { account in
```
utkrishtsahu: This bypasses everything clientProvider sets up (lines 55–65): no .using(inLibrary:version:) and no useDPoP(). So loginWithOTP and multifactorChallenge now go out without the SDK user-agent header, and ignore useDPoP even when the caller requested it. Apply the same using(...)/DPoP setup to the MFA client — note it needs the userAgent/arguments, which the current closure signature ((_ account: Account)) doesn't take.

auth0_flutter_platform_interface/lib/src/auth/auth_multifactor_challenge_options.dart
```diff
@@ -1,20 +1,13 @@
 import '../request/request_options.dart';
-import './challenge_type.dart';
 
 class AuthMultifactorChallengeOptions implements RequestOptions {
   final String mfaToken;
-  final List<ChallengeType>? types;
-  final String? authenticatorId;
+  final String authenticatorId;
```
utkrishtsahu: authenticatorId is now required and types removed on the shared platform-interface, but only the Swift native side is updated to match. Android's MultifactorChallengeApiRequestHandler.kt still reads request.data["types"] and treats authenticatorId as optional — after this merges, every Android multifactorChallenge call receives types == null and a shape it doesn't expect. The Dart signature change and the Android handler need to move together.

auth0_flutter/darwin/auth0_flutter/Package.swift
```diff
@@ -11,8 +11,8 @@ let package = Package(
         .library(name: "auth0-flutter", targets: ["auth0_flutter"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/auth0/Auth0.swift", exact: "2.23.0"),
-        .package(url: "https://github.com/auth0/JWTDecode.swift", exact: "3.3.0"),
+        .package(url: "https://github.com/auth0/Auth0.swift", from: "3.0.0"),
```
utkrishtsahu: SPM floats to any 3.x here while the podspecs pin exact 3.0.1 (and the PR title says 3.0.2). SPM and CocoaPods consumers can resolve to different Auth0.swift builds and CI only exercises one. Pick one version and match it across Package.swift + all three podspecs.

auth0_flutter/darwin/auth0_flutter.podspec
```diff
@@ -19,8 +19,8 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.23.0'
-  s.dependency 'JWTDecode', '3.3.0'
+  s.dependency 'Auth0', '3.0.1'
```
utkrishtsahu: Two things on this line: (a) version mismatch with Package.swift's from: "3.0.0" and the 3.0.2 in the PR title — reconcile; (b) the 3.0 bump silently inherits Auth0.swift's new defaults (scope gaining offline_access, minTTL default 0 → 60), which are runtime behavior shifts for existing apps the moment this merges. Confirm intended or gate/document.

auth0_flutter/example/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
utkrishtsahu, line 1: Deleting these means the example apps re-resolve to whatever 3.x is newest on each build — non-reproducible. Regenerate pinned to the chosen version, or gitignore them deliberately.

utkrishtsahu (changes requested):

auth0_flutter/example/ios/Runner.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
NandanPrabhu: reverted this @utkrishtsahu. Removing this wont make much impact as this on example app but reverted to keep the PR scope confined

auth0_flutter/darwin/auth0_flutter.podspec
NandanPrabhu: good catch. for flutter we are maintaining exact version match. Version must be 3.0.2

auth0_flutter/darwin/auth0_flutter/Package.swift
NandanPrabhu: good catch. for flutter we are maintaining exact version match. Version must be 3.0.2

auth0_flutter_platform_interface/lib/src/auth/auth_multifactor_challenge_options.dart
NandanPrabhu: this is used in deprecated mfa present in authenticationclient which was removed in auth0.android v4. Will revert the change on dart side but this will go away when we will removed authclient mfa apis altogether as part of v3

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/CredentialsManager/CredentialsManagerClearMethodHandler.swift
NandanPrabhu: error returned by catch is returned as part of flutter result

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/CredentialsManager/CredentialsManagerSaveMethodHandler.swift
NandanPrabhu: error thrown by try catch flow will be returned as part of flutter result

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/AuthAPI/AuthAPIHandler.swift
NandanPrabhu: good catch fixed

auth0_flutter/macos/auth0_flutter.podspec
```diff
@@ -19,8 +19,8 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.23.0'
-  s.dependency 'JWTDecode', '3.3.0'
+  s.dependency 'Auth0', '3.0.1'
```
utkrishtsahu: Same as the iOS podspec — still on 3.0.1 while darwin/Package.swift are 3.0.2. Bump to 3.0.2.

auth0_flutter/ios/auth0_flutter.podspec
```diff
@@ -19,8 +19,8 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '11.0'
   s.osx.dependency 'FlutterMacOS'
 
-  s.dependency 'Auth0', '2.23.0'
-  s.dependency 'JWTDecode', '3.3.0'
+  s.dependency 'Auth0', '3.0.1'
```
utkrishtsahu: this is pinned to 3.0.1, but darwin/auth0_flutter.podspec:22 and Package.swift:14 are on 3.0.2. Since all three podspecs are live (each has its own Classes/), CocoaPods consumers can resolve a different Auth0.swift build than SPM. Bump to 3.0.2 to match.

utkrishtsahu (changes requested):

auth0_flutter/ios/auth0_flutter.podspec
NandanPrabhu: updated the versions to match the same exact version across podspec and Package.swift

utkrishtsahu (approved):

auth0_flutter/macos/auth0_flutter.podspec
NandanPrabhu: updated the version to match the same exact one used across pubspec and package.swift

PR #924, opened 2026-08-20
 add ephemeral sessions and registerCallbacks for Android web auth

utkrishtsahu:

### 📋 Changes

Implements on the Android v3 track: ephemeral Web Auth sessions and lifecycle-safe login-result recovery for Auth0.Android v4. Targets `develop/v3.0`.

**Ephemeral sessions** (`LoginWebAuthRequestHandler`)
- Wires the existing (previously iOS/macOS-only) `useEphemeralSession` Web Auth login option to Android via `WebAuthProvider.Builder.withEphemeralBrowsing()`. When `useEphemeralSession: true`, the browser uses a private/ephemeral session that isn't persisted.
- No public API change — the `useEphemeralSession` parameter already exists in the shared `platform_interface`; Android simply now honors it.

**Lifecycle-safe callbacks** (`Auth0FlutterPlugin`)
- Migrates the deprecated `WebAuthProvider.addCallback()` / `removeCallback()` pattern to v4's `WebAuthProvider.registerCallbacks(lifecycleOwner, loginCallback, logoutCallback)`.
- The `LifecycleOwner` is obtained via `FlutterLifecycleAdapter.getActivityLifecycle(binding)`; `registerCallbacks` auto-removes the callback when the activity is destroyed, so the explicit removals in `onDetachedFromActivity*` are dropped.
- Makes login-result delivery survive configuration changes / process death (fixing the rotation bug class). Logout recovery is not surfaced to Dart (the Dart layer only listens for recovered login results), so a no-op logout callback is passed to satisfy the new signature — preserving pre-v4 behavior.
- No public/Dart API change.

### 🎯 Testing

**Unit tests (automated):**
- `LoginWebAuthRequestHandlerTest` — added tests asserting `withEphemeralBrowsing()` is invoked when `useEphemeralSession` is `true`, and not invoked when `false`/absent.
- `Auth0FlutterPluginTest` — updated the `onAttachedToActivity` test to stub the activity lifecycle (`HiddenLifecycleReference(mock<Lifecycle>())`) so the new `registerCallbacks` path is exercised without error.


utkrishtsahu:
Split into two focused PRs per review: #927 (ephemeral `useEphemeralSession` support) and #928 (lifecycle-aware `registerCallbacks`). Closing this combined PR.

PR #927, opened 2026-08-21
feat: honor useEphemeralSession on Android web auth

utkrishtsahu:
### Changes
- `LoginWebAuthRequestHandler` now calls `withEphemeralBrowsing()` when the `useEphemeralSession` login option is `true`, so Android honors it via Auth0.Android v4's ephemeral browsing (matching iOS/macOS). Previously it was accepted but ignored on Android.
- Adds unit tests covering `useEphemeralSession` true / false / absent.
- Documents the behavior change in `V3_MIGRATION_GUIDE.md`.

### Testing
- `./gradlew testDebugUnitTest` (new ephemeral tests pass).

### DIFF
--- a/auth0_flutter/V3_MIGRATION_GUIDE.md
+++ b/auth0_flutter/V3_MIGRATION_GUIDE.md
@@ -127,6 +127,11 @@ user is not silently signed in from an existing browser session.
 await auth0.webAuthentication().login(useEphemeralSession: true);
 ```
 
+> **Note**
+> Ephemeral sessions depend on the device browser supporting them. If the
+> browser that handles the login does not support ephemeral browsing, Android
+> falls back to a normal Custom Tabs login and the session is not ephemeral.
+
 ## Getting Help
 
 If you encounter issues migrating, please open an issue on the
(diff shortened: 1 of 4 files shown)


auth0_flutter/V3_MIGRATION_GUIDE.md
```diff
@@ -106,6 +107,26 @@ final challenge = await auth0.api.multifactorChallenge(
...
+expecting it to be ignored, the login now runs in a private/ephemeral browser
+session: no session is shared with or persisted in the system browser, so the
+user is not silently signed in from an existing browser session.
+
+**Migration:** No code change is required. Review your use of
+`useEphemeralSession` on Android if you relied on the previous no-op behavior:
+
+```dart
+// v3: on Android this now starts an ephemeral browser session (as on iOS/macOS).
+await auth0.webAuthentication().login(useEphemeralSession: true);
```
pmathew92, line 127: Callout that on browsers ephemeral session is not supported, normal custom tab login will happen.

utkrishtsahu: added a note under the code sample: ephemeral sessions depend on the device browser supporting them, and if it doesn't, Android falls back to a normal Custom Tabs login (session is not ephemeral).

pmathew92: This note should be called out in the Examples.md file too if we have a section for epheremal browsing

utkrishtsahu: added a note in example.md

pmathew92 (approved):

PR #928, opened 2026-08-21
fix: register Android web auth callbacks with the activity lifecycle

utkrishtsahu:
### Changes
- Replaces the deprecated global `WebAuthProvider.addCallback`/`removeCallback` with the lifecycle-aware `WebAuthProvider.registerCallbacks(lifecycleOwner, loginCallback, logoutCallback)`, using the activity lifecycle obtained from the plugin binding.
- Login results are now recovered across configuration changes and process death; lifecycle-scoped registration removes the manual add/remove in the detach callbacks.
- Adds a lifecycle stub to the existing `onAttachedToActivity` test.
- Documents the internal improvement in `V3_MIGRATION_GUIDE.md`.

### Testing
- `./gradlew testDebugUnitTest` (plugin tests pass).

### DIFF
--- a/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterPlugin.kt
+++ b/auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterPlugin.kt
@@ -1,6 +1,8 @@
 package com.auth0.auth0_flutter
 
 import androidx.annotation.NonNull
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
 import com.auth0.android.authentication.AuthenticationException
 import com.auth0.android.callback.Callback
 import com.auth0.android.provider.WebAuthProvider
@@ -12,6 +14,7 @@ import com.auth0.auth0_flutter.request_handlers.my_account.*
 import com.auth0.auth0_flutter.request_handlers.web_auth.LoginWebAuthRequestHandler
 import com.auth0.auth0_flutter.request_handlers.web_auth.LogoutWebAuthRequestHandler
 import io.flutter.embedding.engine.plugins.FlutterPlugin
+import io.flutter.embedding.engine.plugins.lifecycle.HiddenLifecycleReference
 import io.flutter.embedding.engine.plugins.activity.ActivityAware
 import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding
 import io.flutter.plugin.common.MethodCall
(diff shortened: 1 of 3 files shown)


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterPlugin.kt
```diff
@@ -189,20 +197,30 @@ class Auth0FlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware {
   override fun onAttachedToActivity(binding: ActivityPluginBinding) {
     webAuthCallHandler.activity = binding.activity
     credentialsManagerCallHandler.activity = binding.activity
-    WebAuthProvider.addCallback(processDeathCallback)
+    registerWebAuthCallbacks(binding)
   }
 
   override fun onDetachedFromActivityForConfigChanges() {
```
pmathew92, line 203: Can delete this empty method if not being used

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterPlugin.kt
```diff
@@ -189,20 +197,30 @@ class Auth0FlutterPlugin: FlutterPlugin, MethodCallHandler, ActivityAware {
...
-    WebAuthProvider.removeCallback(processDeathCallback)
   }
 
   override fun onReattachedToActivityForConfigChanges(binding: ActivityPluginBinding) {
     webAuthCallHandler.activity = binding.activity
     credentialsManagerCallHandler.activity = binding.activity
-    WebAuthProvider.addCallback(processDeathCallback)
+    registerWebAuthCallbacks(binding)
```
pmathew92, line 209: Do we need to add it again in `onReattachedToActivityForConfigChanges` . Wouldn't this add a duplicate entry , since we are not removing this anywhere ?

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterPlugin.kt
utkrishtsahu: Can't delete it — all four ActivityAware methods are abstract, so it must be overridden or it won't compile. It's empty on purpose: registerCallbacks(lifecycleOwner, …) auto-removes the callback on ON_DESTROY, so the old manual removeCallback isn't needed.

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/Auth0FlutterPlugin.kt
utkrishtsahu: No duplicate — it's a single shared processDeathCallback instance and Auth0.Android keeps callbacks in a CopyOnWriteArraySet, so re-adding it is deduped. It also auto-removes on the old activity's ON_DESTROY. Re-registering here is needed because the new activity has a new lifecycle; without it, recovery breaks after a config change.

pmathew92 (dismissed):

pmathew92 (approved):

PR #935, opened 2026-09-01
feat: reconcile WebAuthenticationException error codes across Android and iOS

utkrishtsahu:

### 📋 Changes

Moves the DPoP (Demonstrating Proof-of-Possession) toggle for Web Authentication from a per-`login()` call parameter to the `Auth0` client constructor, so it is configured once at initialization and applied internally by the SDK on every Web Auth login. This aligns Web Auth with `CredentialsManager` and `Passwordless`, which already consume the init-level `useDPoP` flag.

**Public API changes:**

- `WebAuthentication.login()` — **removed** the `useDPoP` parameter. DPoP is no longer configured per login call.
- `Auth0(...)` constructor — the existing `useDPoP` flag now also governs Web Authentication. It is threaded through `Auth0.webAuthentication()` into `WebAuthentication` and applied internally on each `login()`.
- `WebAuthentication` constructor — now accepts an internal `useDPoP` flag (set by `Auth0.webAuthentication()`; not intended for direct instantiation).

**Usage:**

```dart
// Configure DPoP once at init; the SDK applies it to every Web Auth login.
final auth0 = Auth0('DOMAIN', 'CLIENT_ID', useDPoP: true);

final credentials = await auth0.webAuthentication().login();
```

`useDPoP` defaults to `false` (Bearer tokens). When omitted, behavior is unchanged.

### 🎯 Testing

- Unit tests in `auth0_flutter/test/mobile/web_authentication_test.dart` were updated to set `useDPoP: true` on the `Auth0(...)` constructor instead of on `.login()`. The DPoP Authentication and DPoP Integration test groups assert that the flag set at init reaches the native `WebAuthLoginOptions.useDPoP`, and the "disabled" / "defaults to false" cases assert `useDPoP == false` when the flag is not set. All tests pass:

### DIFF
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -416,15 +416,19 @@ try {
   final credentials = await auth0.webAuthentication().login();
   // ...
 } on WebAuthenticationException catch (e) {
-  if (e.isRetryable) {
+  if (e.isUserCancelledException) {
+    // User dismissed the browser — same on Android and iOS
+  } else if (e.isRetryable) {
     // Transient error (e.g. network issue) — safe to retry
+  } else if (e.code == 'dpop_jkt_mismatch') {
+    // DPoP thumbprint mismatch returned by the Auth0 server (iOS)
   } else {
     print(e);
   }
 }
 ```
 
-The `isRetryable` property indicates whether the error is transient (e.g. a network outage) and the operation can be retried.
+`isUserCancelledException` returns `true` when the user dismisses the browser without completing login — on both Android and iOS. The `isRetryable` property indicates whether the error is transient (e.g. a network outage) and the operation can be retried. For server-returned errors (e.g. `dpop_jkt_mismatch`, `access_denied`), `exception.code` contains the raw server error code directly.
 
 </details>
 
--- a/auth0_flutter/EXAMPLES.md
+++ b/auth0_flutter/EXAMPLES.md
@@ -421,14 +421,16 @@ try {
   } else if (e.isRetryable) {
     // Transient error (e.g. network issue) — safe to retry
   } else if (e.code == 'dpop_jkt_mismatch') {
-    // DPoP thumbprint mismatch returned by the Auth0 server (iOS)
+    // DPoP thumbprint mismatch returned by the Auth0 server (Android + iOS)
   } else {
     print(e);
   }
 }
 ```
 
-`isUserCancelledException` returns `true` when the user dismisses the browser without completing login — on both Android and iOS. The `isRetryable` property indicates whether the error is transient (e.g. a network outage) and the operation can be retried. For server-returned errors (e.g. `dpop_jkt_mismatch`, `access_denied`), `exception.code` contains the raw server error code directly.
+- `isUserCancelledException` — `true` when the user dismisses the browser without completing login (Android + iOS).
+- `isRetryable` — `true` when the error is transient (e.g. a network outage) and the operation can be retried.
+- For server-returned errors (e.g. `dpop_jkt_mismatch`, `access_denied`), `exception.code` contains the raw server error code directly.
 
 </details>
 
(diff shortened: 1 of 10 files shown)


auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/WebAuthExceptionExtensions.kt
```diff
@@ -0,0 +1,17 @@
...
+import io.flutter.plugin.common.MethodChannel
+
+fun AuthenticationException.toWebAuthResult(result: MethodChannel.Result) {
+    val details = mutableMapOf<String, Any>("_isRetryable" to isNetworkError)
+    cause?.let {
+        details["cause"] = it.toString()
+        details["causeStackTrace"] = it.stackTraceToString()
+    }
+    val code = when {
+        isCanceled -> "USER_CANCELLED"
```
pmathew92, line 13: Is this discrepancy only for this error code ?

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/WebAuth/WebAuthExtensions.swift
```diff
@@ -8,15 +8,18 @@ import FlutterMacOS
 
 extension FlutterError {
     convenience init(from webAuthError: WebAuthError) {
+        func causeCode(fallback: String) -> String {
```
pmathew92, line 11: What is this method doing and when do we use this ?

auth0_flutter/EXAMPLES.md
```diff
@@ -416,15 +416,19 @@ try {
   final credentials = await auth0.webAuthentication().login();
   // ...
 } on WebAuthenticationException catch (e) {
-  if (e.isRetryable) {
+  if (e.isUserCancelledException) {
+    // User dismissed the browser — same on Android and iOS
+  } else if (e.isRetryable) {
     // Transient error (e.g. network issue) — safe to retry
+  } else if (e.code == 'dpop_jkt_mismatch') {
+    // DPoP thumbprint mismatch returned by the Auth0 server (iOS)
```
pmathew92: Why is this error mapped only for iOS ? Android also would be throwing similar error

auth0_flutter/EXAMPLES.md
```diff
@@ -416,15 +416,19 @@ try {
...
   } else {
     print(e);
   }
 }
 ```
 
-The `isRetryable` property indicates whether the error is transient (e.g. a network outage) and the operation can be retried.
+`isUserCancelledException` returns `true` when the user dismisses the browser without completing login — on both Android and iOS. The `isRetryable` property indicates whether the error is transient (e.g. a network outage) and the operation can be retried. For server-returned errors (e.g. `dpop_jkt_mismatch`, `access_denied`), `exception.code` contains the raw server error code directly.
```
pmathew92: Lets note this down as bullet points

auth0_flutter/android/src/main/kotlin/com/auth0/auth0_flutter/WebAuthExceptionExtensions.kt
utkrishtsahu: Yes. Cancel was the only code that differed — Android emitted a0.authentication_canceled vs iOS USER_CANCELLED. So we only normalize that one case; every other code is passed straight through via getCode() unchanged.

auth0_flutter/darwin/auth0_flutter/Sources/auth0_flutter/WebAuth/WebAuthExtensions.swift
utkrishtsahu: causeCode(fallback:) unwraps the underlying server error. In Auth0.swift v3, server errors like dpop_jkt_mismatch, invalid_grant, access_denied are wrapped inside authenticationFailed / codeExchangeFailed / other as an AuthenticationError cause. This extracts that cause's .code so the real server code surfaces as exception.code; if the cause isn't an AuthenticationError, it falls back to the generic wrapper name (e.g. AUTHENTICATION_FAILED).

auth0_flutter/EXAMPLES.md
utkrishtsahu: corrected to (Android + iOS). Both platforms now surface the raw server error code directly. Fixed in the latest commit.

auth0_flutter/EXAMPLES.md
utkrishtsahu: Done — reformatted into bullet points. Fixed in the latest commit.

pmathew92 (approved):

